PR #2, opened 2018-07-29
Added pointer dereferencing for field extraction

lmika:
Hi @antonmedv,

Saw this on Golang Weekly and though it looked pretty good.  Added a PR which dereferences pointers to structs when attempting to extract a field name so that pointers to structs can be used as data alongside regular structs.  Example:

```!go
type Data struct {
   X int
}

data := &Data{X: 123}
result, err := expr.Eval("X", data)
// result == 123
```

Hope you find it useful.

Thanks,

Leon

### DIFF
--- a/eval_test.go
+++ b/eval_test.go
@@ -130,6 +130,11 @@ var evalTests = []evalTest{
 		true,
 	},
 	{
+		`Foo.Bar`,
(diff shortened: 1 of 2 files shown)


antonmedv:
Really cool! Thanks!

PR #16, opened 2018-11-10
Match string in string

relvacode:
Adds ability to match if `"foo" in "foobar"`

### DIFF
--- a/eval_test.go
+++ b/eval_test.go
@@ -275,6 +275,16 @@ var evalTests = []evalTest{
 		false,
 	},
 	{
+		`"bar" in "foobar"`,
+		nil,
+		true,
+	},
+	{
+		`"bar" in "world"`,
+		nil,
+		false,
+	},
+	{
 		`A == nil`,
 		struct{ A interface{} }{nil},
 		true,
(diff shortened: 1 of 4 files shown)


antonmedv:
But we have _matches_.  It does same think:

```ruby
"foo" matches "foobar"
```

Is it?

relvacode:
These examples don't work in the same way as they would with `in` as they do with a regular expression match. To me they seem different enough.

```
"foobar" matches "foo"  (should be false)
"foo" matches "foobar"  (should be true)
"foobar" matches ""     (should be false)
"$1" matches "$1.50"    (should be true)
"foo" matches "foo.bar" (should be true)
```

antonmedv:
I see, not `in` operator supports both maps and arrays. Do you think it will be okay to override strings too?

relvacode:
I don't see why not. It works for me, would you like me to add tests for that?

antonmedv:
Ok, just please add type checks for strict mode. (See type.go)

relvacode:
Type checks added

antonmedv:
After discussion with my team we decided not to add this functionality as it can be error prone.
For example `"" in bar` if user expects bar to be an array he may check for empty string properly, but if bar ended up being string this check will always be try. 
(In fact we see this errors in python)

As alternative is to define helper method:

```ruby
Contains("foobar", "bar")
``` 

What do you think?

antonmedv:
I'm working on next expr version and it will have `contains` operator.

antonmedv:
Will be solved in #21

PR #17, opened 2018-12-06
Add basic visitor support

etcinitd:
**Not ready to merge yet!**
Hi @antonmedv 
I needed a feature that allows me to walk the parsed AST (without evaluation) to find variables and functions used in an expression. This was my first try that I wanted to get your opinions on. 
What do you think? Is it the right approach? How can I make it good enough to be merged into the upstream master?
Cheers,
Amir

### DIFF
--- a/bench_test.go
+++ b/bench_test.go
@@ -3,7 +3,7 @@ package expr_test
 import (
 	"testing"
 
-	"github.com/antonmedv/expr"
+	"github.com/etcinitd/expr"
 )
 
 type Segment struct {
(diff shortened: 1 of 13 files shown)


antonmedv:
I was thinking about adding visitor myself. But still haven't found time for this. Will be cool to see API there custom visitor can define only needed nodes wishing to visit. 

See babel for example, pretty cool  implementation.

etcinitd:
Then we need some form of defining node types, I guess. For example, we can make node structs exported. Any other ideas?

antonmedv:
If visitor will be added, AST nodes must be exported.

antonmedv:
Will be solved in #21

PR #19, opened 2019-01-19
Add support for method of map.

szyhf:
Hi, I find out that while calling `getFunc(...)`, when the kind of val is `Map`, will only assume the method is provide by k-v. But even a map can have it is own method, and it will help a lot. 
For example:

```go
type ExprParams map[string]interface{}

func (e ExprParams) Max(a, b float64) float64 {
	return math.Max(a, b)
}

type User struct{
    Age int
}

func demo(){
    expr.Eval("Max(Teacher.Age, Student.Age)",ExprParams{
		"Teacher":User{Age:28},
		"Student":Student{Age:8},
	})
}

```

In this cases, `ExprParams` can provide default method ( like `Max`) easily.

And the implement is very simple, I hope it can improve the lib.

Thanks a lot.

PS: another question
I find that the only to calling for `getFunc(val interface{}, i interface{})`, the param `i` is both `string`, why not just change the `getFunc(...)` to `getFunc(val interface{}, funcName string)` ? That will save the cost of `reflect.ValueOf(i).String()`

### DIFF
--- a/eval_test.go
+++ b/eval_test.go
@@ -21,6 +21,22 @@ type evalErrorTest struct {
 	err   string
 }
 
+type evalParams map[string]interface{}
+
+func (p evalParams) Max(a, b float64) float64 {
+	if a < b {
+		return b
+	}
+	return a
+}
+
+func (p evalParams) Min(a, b float64) float64 {
+	if a < b {
+		return a
+	}
+	return b
+}
+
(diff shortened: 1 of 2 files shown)


antonmedv:
Awesome! Can you add tests for this case?

szyhf:
Sure, but where may I put the test case ?

PR #31, opened 2019-06-16
Allow overloading binary node with a function from env

davidklassen:
resolves https://github.com/antonmedv/expr/issues/27

### DIFF
--- a/cmd/exe/debugger.go
+++ b/cmd/exe/debugger.go
@@ -7,6 +7,7 @@ import (
 	. "github.com/antonmedv/expr/vm"
 	"github.com/gdamore/tcell"
 	"github.com/rivo/tview"
+	"github.com/sanity-io/litter"
 	"sort"
 	"strconv"
 	"strings"
@@ -21,7 +22,6 @@ func debugger() {
 	check(err)
 
 	vm := NewVM(true)
-	go vm.Run(program, nil)
 
 	app := tview.NewApplication()
 	table := tview.NewTable()
@@ -41,6 +41,19 @@ func debugger() {
 		AddItem(table, 0, 1, true).
 		AddItem(sub, 0, 1, false)
 	app.SetRoot(flex, true)
+
+	go func() {
+		out := vm.Run(program, nil)
+		app.QueueUpdateDraw(func() {
+			sub.RemoveItem(scope)
+			result := tview.NewTextView()
+			result.
+				SetBorder(true).
+				SetTitle("Output")
+			result.SetText(litter.Sdump(out))
+			sub.AddItem(result, 0, 1, false)
+		})
+	}()
 
 	index := make(map[int]int)
 	for row, line := range strings.Split(program.Disassemble(), "\n") {


checker/checker.go
```diff
@@ -8,7 +8,7 @@ import (
 	"reflect"
 )
 
-func Check(tree *parser.Tree, types TypesTable) (t reflect.Type, err error) {
+func Check(tree *parser.Tree, types TypesTable, operators map[string][]string) (t reflect.Type, err error) {
```
antonmedv: Let's split type `map[string][]string` to smaller types with comments of what they are containing.

checker/checker.go
```diff
@@ -144,6 +146,27 @@ func (v *visitor) BinaryNode(node *ast.BinaryNode) reflect.Type {
 	l := v.visit(node.Left)
 	r := v.visit(node.Right)
 
+	// check operator overloading
+	if fns, ok := v.operators[node.Operator]; ok {
+		for _, fn := range fns {
+			fnType, ok := v.types[fn]
+			if !ok || fnType.Type.Kind() != reflect.Func {
+				continue
```
antonmedv: What if you specify operator overload on not existing helper's func? Here will be no errors. 
Maybe it's better to check overloads before starting compilation?

checker/checker.go
```diff
@@ -144,6 +146,27 @@ func (v *visitor) BinaryNode(node *ast.BinaryNode) reflect.Type {
...
+			}
+
+			if fnType.Type.NumIn() != 2 || fnType.Type.NumOut() != 1 {
```
antonmedv: Here is again, we can check this before checker phase, in "configuration" phase.
And no need to wait on then a proper expression will be compiled to get an error.

expr.go
```diff
@@ -84,6 +94,55 @@ func CompileType(input string, ops ...OptionFn) (*vm.Program, reflect.Type, erro
 	return program, t, nil
 }
 
+type operatorPatcher struct {
```
antonmedv: Let's move it to another package, this top level package is for reexporting only. Maybe move it to `checker` or to new package (like `postprocessing`)

expr.go
```diff
@@ -84,6 +94,55 @@ func CompileType(input string, ops ...OptionFn) (*vm.Program, reflect.Type, erro
...
+	ast.BaseVisitor
+	ops   map[string][]string
+	types checker.TypesTable
+}
+
+func (p *operatorPatcher) Node(node *ast.Node) {
+	binaryNode, ok := (*node).(*ast.BinaryNode)
```
antonmedv: Why not using `BinaryNode()` method?

expr_test.go
```diff
@@ -218,3 +264,149 @@ func ExampleEval_marshal() {
...
+		{
+			name: "+ operator",
+			code: "1 + 1",
+			want: 2,
+		},
+		{
+			name:    "associativity",
+			code:    "(A * B) * C == A * (B * C)",
+			env:     struct{ A, B, C int }{},
+			request: struct{ A, B, C int }{A: 1, B: 2, C: 3},
```
antonmedv: Here is to much boilerplates to write. What about defining it only once and just checking expressions? Take a llot at `vm` package tests.

expr_test.go
```diff
@@ -218,3 +264,149 @@ func ExampleEval_marshal() {
...
+			code:    "A[0] + A[1]",
+			env:     struct{ A []int }{},
+			request: struct{ A []int }{A: []int{1, 2}},
+			want:    3,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			program, err := expr.Compile(tt.code, expr.Env(tt.env))
+			if err != nil {
```
antonmedv: Let's use `assert`/`require` as other tests in expr.

PR #34, opened 2019-06-23
Allow overloading binary node with a function from env

davidklassen:
fixes https://github.com/antonmedv/expr/issues/27

### DIFF
--- a/checker/patcher.go
+++ b/checker/patcher.go
@@ -1,0 +1,49 @@
+package checker
+
+import (
+	"github.com/antonmedv/expr/ast"
+	"github.com/antonmedv/expr/internal/conf"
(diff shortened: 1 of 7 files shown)


checker/patcher.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+type operatorPatcher struct {
+	ast.BaseVisitor
+	ops   map[string][]string
+	types conf.TypesTable
+}
+
+func (p *operatorPatcher) Node(node *ast.Node) {
+	binaryNode, ok := (*node).(*ast.BinaryNode)
```
davidklassen: Have to use generic node here to be able to replace it in the tree

Issue #36, opened 2019-07-03
Compile of method call on interface fails with "too many arguments to call Method (1:1)"


enijns:
Using expr.Eval seems to work fine though.
I will also submit the following example as a pull request:

`type AnInterface interface {
	Method(x string) []AnInterface
}

type A struct{ I int }

func (self A) Method(x string) []AnInterface {
	return []AnInterface{self}
}

func ExampleCompileInterfaceMethodCall() {

	type B_Env struct{ Ai AnInterface }

	env := B_Env{A{6}}

	prog, err := expr.Compile(`Ai.Method("x")`, expr.Env(env))

	if err != nil {
		fmt.Printf("%v", err)
		return
	}

	output, err := expr.Run(prog, env)

	if err != nil {
		fmt.Printf("%v", err)
		return
	}

	fmt.Printf("%v", output)

	// Output: 6
}`

antonmedv:
Fixed in 9c096dbcfb99ec4f3f42a846c34e86864c062145

PR #37, opened 2019-07-03
Compile of method call on interface fails with "too many arguments to call Method"

enijns:
see #36


antonmedv:
I see. Problem with the checker.

antonmedv:
Fixed in 9c096dbcfb99ec4f3f42a846c34e86864c062145

enijns:
Thank you for the quick resolution!

PR #44, opened 2019-07-20
fix map [string] interface{}  is nil

lian-yue:
```go
	boolVal := false
	if out, err = expr.Eval("f != nil", map[string]interface{}{"f": boolVal}); err != nil {
		panic(err)
	}
	fmt.Println("out", out)


	trueVal := false
	if out, err = expr.Eval("f == true", map[string]interface{}{"f": trueVal}); err != nil {
		panic(err)
	}
	fmt.Println("out", out)


	stringVal := "string"
	if out, err = expr.Eval(`s == "string"`, map[string]interface{}{"s": stringVal}); err != nil {
		panic(err)
	}
	fmt.Println("out", out)


	var timeNil *time.Time
	if out, err = expr.Eval("t == nil", map[string]interface{}{"t": timeNil}); err != nil {
		panic(err)
	}
	fmt.Println("out", out)
```

### DIFF
--- a/vm/generate/main.go
+++ b/vm/generate/main.go
@@ -140,20 +140,6 @@ func main() {
 		echo(``)
 	}
 
-	// isNil func
-	echo(`func isNil(v interface{}) bool {`)
-	echo(`if v == nil {`)
-	echo(`return true`)
-	echo(`}`)
-	echo(`r := reflect.ValueOf(v)`)
-	echo(`switch r.Kind() {`)
-	echo(`case reflect.Chan, reflect.Func, reflect.Map, reflect.Ptr, reflect.Interface, reflect.Slice:`)
-	echo(`return r.IsNil()`)
-	echo(`default:`)
-	echo(`return false`)
-	echo(`}`)
-	echo(`}`)
-
 	b, err := format.Source([]byte(data))
 	check(err)
 	err = ioutil.WriteFile("helpers.go", b, 0644)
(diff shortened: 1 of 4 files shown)


vm/helpers.go
```diff
@@ -332,15 +332,31 @@ func equal(a, b interface{}) interface{} {
...
 	case string:
-		return x == b.(string)
+		switch y := b.(type) {
+		case string:
+			return x == y
+		}
 	}
 	// Two nil values should be considered as equal.
-	if (a == nil || reflect.ValueOf(a).IsNil()) && (b == nil || reflect.ValueOf(b).IsNil()) {
+	if isNil(a) && isNil(b) {
```
antonmedv: This should go into generate helper.

lian-yue: Added to generate helper

lian-yue:
Added to generate helper

vm/generate/main.go
```diff
@@ -137,6 +140,20 @@ func main() {
 		echo(``)
 	}
 
+	// isNil func
+	echo(`func isNil(v interface{}) bool {`)
```
antonmedv: Better put it in runtime.go file.

lian-yue:
ok

antonmedv:
Fixed in 96a4cee7a22c684d5dbe44fdb0339f3270ce884d

PR #49, opened 2019-08-14
add support for environment lookup function

bontibon:


antonmedv:
What's the purpose of this PR? Why can't you use map[string]interface{}?

PR #54, opened 2019-08-17
runtime: Don't value copy structs from pointers on property access

deanm:
Adds a test with a 10mb struct passed by a pointer, it would be previously
copied on property accessed whereas now the accesses just go through the
pointer without a copy.

Previous:
Benchmark_largeStructAccess-4          200     7914350 ns/op

After:
Benchmark_largeStructAccess-4      2000000         912 ns/op

### DIFF
--- a/bench_test.go
+++ b/bench_test.go
@@ -97,3 +97,25 @@ func Benchmark_call(b *testing.B) {
 		b.Fatal(err)
 	}
 }
+
+func Benchmark_largeStructAccess(b *testing.B) {
+	type Env struct {
+		Data [1024*1024*10]byte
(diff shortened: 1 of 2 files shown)


antonmedv:
Thanks! Awesome contribution! Maybe you can leave some testimonials in README's section "Who is using Expr?"?

deanm:
> Thanks! Awesome contribution! Maybe you can leave some testimonials in README's section "Who is using Expr?"?

Not using it yet : )

antonmedv:
Nice catch :)

Well, if you are going to use it some there :)

PR #57, opened 2019-08-19
vm: Use a preallocated fixed call frame for typical argument sizes

deanm:
For calls with 20 or less arguments use a preallocated fixed array
for the calling frame, otherwise fall back to heap allocation. This
avoids what was previously always a heap allocation at callsites.

Before:
Benchmark_Calls-4        50000       27618 ns/op

After:
Benchmark_Calls-4        50000       25948 ns/op

### DIFF
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -54,6 +54,13 @@ func (vm *VM) Run(program *Program, env interface{}) interface{} {
 	vm.bytecode = program.Bytecode
 	vm.constants = program.Constants
 
+	// Avoid having to dynamically allocate the calling arguments frame in the
+	// typical case of calls, if larger number of arguments then will alloc.
+	// This is possible because the vm never calls back into itself / nests calls
+	// (expr code cannot call into a function defined in expr code), so the
+	// callstack is only ever a single frame.
+	var fixedFrame20 [20]reflect.Value
+
 	for vm.ip < len(vm.bytecode) {
 
 		if vm.debug {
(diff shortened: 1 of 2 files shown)


XCapsule:
approve

savsgio:
any progress or news?

antonmedv:
@savsgio A little bit old PR. Looks this PR needs to be rebased. What is your use case? I think even without VM reuse Exor Is pretty fast.

vm/vm.go
```diff
@@ -50,10 +38,41 @@ func NewVM(debug bool) *VM {
...
 func (vm *VM) Run(program *Program, env interface{}) interface{} {
 	vm.bytecode = program.Bytecode
 	vm.constants = program.Constants
 
+	// Avoid having to dynamically allocate the calling arguments frame in the
+	// typical case of calls, if larger number of arguments then will alloc.
+	// This is possible because the vm never calls back into itself / nests calls
+	// (expr code cannot call into a function defined in expr code), so the
+	// callstack is only ever a single frame.
+	var fixedFrame20 [20]reflect.Value
```
antonmedv, line 74: How does it affect expressions without calls at all? We gonna allocate frames for every VM run. Even there is no calls. Mabe this part should be separated in another PR? And added compile time detection if it's needed to allocate fixedFrame20. Also is there any performance boost from it?

savsgio:
> @savsgio A little bit old PR. Looks this PR needs to be rebased. What is your use case? I think even without VM reuse Exor Is pretty fast.

I need to reuse the same VM in multiples executions. Now i need to evaluate some rules per each request so if i use `vm.Run(...)` that's generate one memory allocation (the VM generation), that reduce the performance significantly.
 
I tried to use, but form seconde execution, always returns `nil`

```go
rule, _ := expr.Compile("'foo' == 'foo'")
bar := vm.NewVM(false)

fmt.Println(bar.Run(rule, nil)) // true
fmt.Println(bar.Run(rule, nil)) // nil
// the next n executions always returns nil
``

**Comment by savsgio** [comment]
Reducing this allocation, will increase the performance.
```
Benchmark_Expr-4   	12952581	        92.7 ns/op	      32 B/op	       1 allocs/op
```

The code of this benchmark:
```go
func Benchmark_Expr(b *testing.B) {
	foo, _ := expr.Compile("'foo' == 'foo'")

	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		vm.Run(foo, nil)
	}
}
```

antonmedv:
Okay, can you make new PR based on this?

antonmedv:
Did it here: #107

savsgio:
Nice!

PR #60, opened 2019-08-19
vm: Add a zerocopy mode for avoiding value copies.

deanm:
This is not backwards compatible, since arguments to functions can change from
having been a type to a *type now, and there isn't an easy way to track and fix
that up.  However, callers can opt-in to that behaviour with SetZeroCopy().

### DIFF
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -36,6 +36,7 @@ type VM struct {
 	debug     bool
 	step      chan struct{}
 	curr      chan int
+	zerocopy  bool
 }
 
 func NewVM(debug bool) *VM {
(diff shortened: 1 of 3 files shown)


vm/vm.go
```diff
@@ -36,6 +36,7 @@ type VM struct {
 	debug     bool
 	step      chan struct{}
 	curr      chan int
+	zerocopy	bool
```
bontibon: `go fmt`

deanm: fixed

antonmedv:
This is really not trivial DX. Maybe take a look at using reflect.Value instead of interface{}?

deanm:
I agree, but I won't have more time to work on it, so this is just the quick solution I came up with for now.  At least you now know of the problems and possible solutions.

Thanks

antonmedv:
Well, after investigation I think is better just to use pointers and slices :)

```
Benchmark_largeStructAccess-8         	 1344613	       863 ns/op
Benchmark_largeNestedStructAccess-8   	 1097618	      1114 ns/op
Benchmark_largeNestedArrayAccess-8    	 2307724	       510 ns/op
```

```go
	type I struct {
		Data [1024*1024*10]byte
		Field int
	}
	type Env struct {
		Inner *I
	}
```

This way only pointer is copied. 

Also ther is a way better performance can be achieved in case of value types — store reflect.Value on stack itself, and add opcode to `.Interface()` it.

PR #72, opened 2019-10-10
Dereference pointer types

Dasio:


antonmedv:
Hi, can you show an example and test case there this is happening? Maybe we can put dereferencing on `runtime.fetch`?

Dasio:
```go
package main

import (
	"fmt"
	"github.com/antonmedv/expr"
)

type Product struct {
	OverallRating *int
}

func main() {
	val := 50
	product := Product{
		OverallRating: &val,
	}
	expression := "OverallRating < 80"
	output, err := expr.Eval(expression, product)
	if err != nil {
		panic(err)
	}
	fmt.Println(output)
}
```
This is use-case which I needed

Dasio:
Instead I modified fetch inside runtime, if field inside struct is pointer, dereference it.

Dasio:
Anyway Test failed - TestOperator_interface. 
 - we would appreciate this feature, if it could be implemented properly.

antonmedv:
Actually, in Go this is also invalid operation:
```
val := 50
var i *int = &val
_ := i > val
```
`./expr_test.go:834:9: invalid operation: i > val (mismatched types *int and int)`

You can create a function for dereferencing:
```go
	val := 50
	product := struct {
		OverallRating *int
		Dereference   func(*int) int
	}{
		OverallRating: &val,
		Dereference: func(i *int) int {
			return *i
		},
	}

	output, err := expr.Eval("Dereference(OverallRating) < 80", product)
```

Or you can override operators for supporting it:

```go
	type Product struct {
		OverallRating *int
		Compare       func(a *int, b int) bool
	}

	val := 50
	product := Product{
		OverallRating: &val,
		Compare: func(a *int, b int) bool {
			return *a < b
		},
	}

	expression := "OverallRating < 80"
	program, err := expr.Compile(expression, expr.Env(Product{}), expr.Operator("<", "Compare"))
	require.NoError(t, err)

	output, err := expr.Run(program, product)
	require.NoError(t, err)
	require.Equal(t, true, output)
```

Issue #52, opened 2019-08-14
JSON marshaling of Program is incomplete


bontibon:
[The documentation states](https://github.com/antonmedv/expr/blob/master/docs/Usage.md#marshaling-program) that marshaling and unmarshaling of a program is supported.
However, the following test fails:

```go
func Test_marshalRegexp(t *testing.T) {
	prog, err := expr.Compile(`"hello" matches "h.*"`)
	if !assert.NoError(t, err) {
		t.FailNow()
	}

	marshaled, err := json.Marshal(prog)
	if !assert.NoError(t, err) {
		t.FailNow()
	}

	unmarshaledProgram := &vm.Program{}
	err = json.Unmarshal(marshaled, unmarshaledProgram)
	if !assert.NoError(t, err) {
		t.FailNow()
	}

	output, err := expr.Run(unmarshaledProgram, nil)
	if !assert.NoError(t, err) {
		t.FailNow()
	}
	assert.Equal(t, true, output)
}
```

antonmedv:
Hmm, I think this is due to cached regex.

rucciva:
hi, i'am interested in using this library in my project, but it requires to store the expression into database.

i've created some benchmark regarding the marshalling process with comparison to govaluate performace and found that Unmarshaling from json takes as much time as re-compiling the script (though it allocated less bytes). Shouldn't vm.Program implement encoding.TextMarshaler instead and use faster serialisation method?

```go
func BenchmarkLibrary(b *testing.B) {
	parameters := map[string]interface{}{
		"s_age": 10,
	}
	s := `s_age == 10 || s_age < 10`

	b.Run("expr", func(b *testing.B) {
		for n := 0; n < b.N; n++ {
			p, _ := expr.Compile(s, expr.Env(parameters))
			expr.Run(p, expr.Env(parameters))
		}
	})

	b.Run("expr-marshalled", func(b *testing.B) {
		p, _ := expr.Compile(s, expr.Env(parameters))
		j, _ := json.Marshal(p)
		for n := 0; n < b.N; n++ {
			uP := &vm.Program{}
			json.Unmarshal(j, &uP)
			expr.Run(uP, expr.Env(parameters))
		}
	})

	b.Run("govaluate", func(b *testing.B) {
		for n := 0; n < b.N; n++ {
			expression, _ := govaluate.NewEvaluableExpression(s)
			expression.Evaluate(parameters)
		}
	})
}
```

```
[...]

PR #74, opened 2019-10-17
WIP: try marshall

rucciva:
#52


rucciva:
@antonmedv , any feedback?

antonmedv:
This is awesome! So now program goes gob to base64 to json?

antonmedv:
What with performance?

rucciva:
> So now program goes gob to base64 to json

Yup, since expr use a lot of Interface{}, we need to encode type information so that it can be decoded-back correctly, and binary encoder like gob do just that . But since we need string for json, base64 is needed.

>What with performance?

Haven't tested it yet, but the step is similliar with the one i write when i first commented on the issue, just now we used gob instead of go-codec.

rucciva:
What about the *regexp.Regexp wrapping? Is that acceptable?

antonmedv:
I think it’s okay, since we anyway need to compile regex on unmarshal before runtime.

rucciva:
Cool then.

antonmedv:
Can you also add benchmarks to the repo so we can track unmarshal performance over time.

rucciva:
Sure...

rucciva:
hi @antonmedv , based on the benchmark test, gob is not a fast encoded. it doesn't increase significance performance.

I've tried with go-codec's binc, but i'm still stuck at decoding *Regexp back since vm.Program only contains nil interface `Constant`

rucciva:
Btw, if parameters passed when compiling has methods, it needs to be registered to gob for successful marshal.

antonmedv:
This is sad. Let’s think, maybe we can throw away something unnecessary for increasing performance?

rucciva:
I am afraid there is none.

I think we need a better encoding-decoding mechanism in terms of speed. But i am afraid it is harder with said mechanism to unmarshall the type back correctly due to the usage if interface{}

antonmedv:
Another possibility is to split constants into separate list for `[]vm.Call`, `[]regex`, `[]string`, etc. 
But this requires a lot of internal changes. 
Maybe we need to take another way, and safe original source on marshal and recompile it on unmarshal. What do you think @rucciva?

rucciva:
> Maybe we need to take another way, and safe original source on marshal and recompile it on unmarshal.

hmm,, i don't think combining unmarshal and recompile is a good idea, because the point of marshall-unmarshal is to avoid recompile. Or do you mean a different kind of recompile?

btw, i've opened up an [issue](https://github.com/ugorji/go/issues/322) regarding decoding back pointer to interface{}. i've tried with go-binc  and it is 5x faster than recompile (still failed to decode-back pointer of Regexp). But the caveat is that any custom parameters type need to provide [extension](https://godoc.org/github.com/ugorji/go/codec#BytesExt) and expr need to expose function to register said extension to go-binc handler

antonmedv:
I tested _json_ vs  _gob-base64-json`:
```
Benchmark_brg/unmarshall_v1-8         	   20906	     53843 ns/op
Benchmark_brg/unmarshall_v2-8         	   26377	     43177 ns/op
```

I think if we want marshal/unmarshal functionality to work correctly to find a better way of doing it.

rucciva:
hi @antonmedv , sorry didn't catch the notification

>I think if we want marshal/unmarshal functionality to work correctly to find a better way of doing it.

yup, in terms of performance, i think its the binary marshaller that we need to find.  Try replacing gob with go-codec (ignoring regexp and other custom type) and you will get around 4x performance improvement. But the downside is adding custom type will be harder, including custom parameters.  

but in terms of functionality correctness (correctly marshall and then unmarshall), using gob maybe the best way, since it is easier for custom type

PR #79, opened 2019-12-05
add timeout

Ashtonian:
needs testing and maybe some more glue but what do you think about adding a runtime timeout functionality ? I've defaulted it to 30 seconds which may be considered a breaking change :/

### DIFF
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -5,6 +5,7 @@ import (
 	"reflect"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/antonmedv/expr/internal/file"
 )
@@ -22,8 +23,25 @@ func Run(program *Program, env interface{}) (out interface{}, err error) {
 		}
 	}()
 
-	out = vm.Run(program, env)
-	return
+	ch := make(chan interface{}, 1)
+	defer close(ch)
+
+	go func() {
+		ch <- vm.Run(program, env)
+	}()
+
+	timer := time.NewTimer(program.Timeout)
+	defer timer.Stop()
+
+	select {
+	case out = <-ch:
+		return
+	case <-timer.C:
+		// TODO: type error so it can be checked externally
+		err = fmt.Errorf("operation timedout. Current Timeout %v", program.Timeout)
+		return
+	}
+
 }
 
 type VM struct {
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -38,6 +38,7 @@ func Compile(tree *parser.Tree, config *conf.Config) (program *Program, err erro
 	}
 
 	program = &Program{
+		Timeout:   config.Timeout,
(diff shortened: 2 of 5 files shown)


antonmedv:
Hi, but why this is needed? I think it’s impossible to loop forever in expr and program eventually stops.

Ashtonian:
Hi, sorry for the code vomit, this is a great library, thank you for your work.

Admittedly this is because of resources I'm introducing to the environment. This use case is for incoming time series data, and I've attached env functions that fetch aggregated data from an in memory store, if the user were to abuse the calling of these functions it could cause problems for the expr runner. Mainly this is out of extreme paranoia running external code. 

Demo: 
```
package main 

var expressions = []string{
	"runForever()",
}

var environment = map[string]interface{}{
	"runForever":       func() {for {} },
}

func main() {
	for _, input := range expressions {
		program, err := expr.Compile(input, expr.Env(environment))
		if err != nil {
			panic(err)

		output, err := expr.Run(program, environment)
		if err != nil {
			panic(err)
		}

		fmt.Println(output)
	}
}
 ```

File: vm/vm.go [hunk 1]
```diff
@@ -22,8 +23,25 @@ func Run(program *Program, env interface{}) (out interface{}, err error) {
...
+	timer := time.NewTimer(program.Timeout)
+	defer timer.Stop()
+
+	select {
+	case out = <-ch:
+		return
+	case <-timer.C:
+		// TODO: type error so it can be checked externally
+		err = fmt.Errorf("operation timedout. Current Timeout %v", program.Timeout)
+		return
```
antonmedv: But this won't stop goroutine and code will be still working. This kind of early return can be done outsize Expr library same way.

antonmedv:
Functions you exposing to a user should be really saved to use. Consider this:
```
func Run(s string) {
    exec.Command(s)
}
```

> Run("rm -rf /")

Ashtonian:
> Functions you exposing to a user should be really saved to use. Consider this:
> 
> ```
> func Run(s string) {
>     exec.Command(s)
> }
> ```
> 
> > Run("rm -rf /")

yea "safe" here indicating that they won't halt the runtime. The functions don't have access to anything they shouldn't and in this case they are just long running.

vm/vm.go
Ashtonian: Thats a good point, the timeout would likely need to be paired with a `interrupt` channel to properly halt the parser. 

Something like [this](https://github.com/robertkrimen/otto/blob/15f95af6e78dcd2030d8195a138bd88d4f403546/cmpl_evaluate_expression.go#L16 ) at the [root](https://github.com/antonmedv/expr/blob/73253fa3d7f2902c5c3be75a9775b6f1aa47157e/compiler/compiler.go#L115) of the parser.

PR #86, opened 2020-02-19
vm.go: reflect calls with nil parameters fix

wintermute-cds:
golang gotcha: reflect calls with nil params fail without this trick

### DIFF
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -251,7 +251,11 @@ func (vm *VM) Run(program *Program, env interface{}) interface{} {
 			in := make([]reflect.Value, call.Size)
 			for i := call.Size - 1; i >= 0; i-- {
 				popped := vm.pop()
-				in[i] = reflect.ValueOf(&popped).Elem()
+				if popped == nil {
+					in[i] = reflect.ValueOf(&popped).Elem()
+				} else {
+					in[i] = reflect.ValueOf(popped)
+				}
 			}
 			out := fetchFn(env, call.Name).Call(in)
 			vm.push(out[0].Interface())
@@ -270,7 +274,11 @@ func (vm *VM) Run(program *Program, env interface{}) interface{} {
 			in := make([]reflect.Value, call.Size)
 			for i := call.Size - 1; i >= 0; i-- {
 				popped := vm.pop()
-				in[i] = reflect.ValueOf(&popped).Elem()
+				if popped == nil {
+					in[i] = reflect.ValueOf(&popped).Elem()
+				} else {
+					in[i] = reflect.ValueOf(popped)
+				}
 			}
 			out := fetchFn(vm.pop(), call.Name).Call(in)
 			vm.push(out[0].Interface())
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -250,7 +250,8 @@ func (vm *VM) Run(program *Program, env interface{}) interface{} {
 			call := vm.constant().(Call)
 			in := make([]reflect.Value, call.Size)
 			for i := call.Size - 1; i >= 0; i-- {
-				in[i] = reflect.ValueOf(vm.pop())
+				popped := vm.pop()
+				in[i] = reflect.ValueOf(&popped).Elem()
 			}
 			out := fetchFn(env, call.Name).Call(in)
 			vm.push(out[0].Interface())
@@ -268,7 +269,8 @@ func (vm *VM) Run(program *Program, env interface{}) interface{} {
 			call := vm.constants[vm.arg()].(Call)
 			in := make([]reflect.Value, call.Size)
 			for i := call.Size - 1; i >= 0; i-- {
-				in[i] = reflect.ValueOf(vm.pop())
+				popped := vm.pop()
+				in[i] = reflect.ValueOf(&popped).Elem()
 			}
 			out := fetchFn(vm.pop(), call.Name).Call(in)
 			vm.push(out[0].Interface())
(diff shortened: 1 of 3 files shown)


wintermute-cds:
we've based this fix on the discussion at https://groups.google.com/forum/#!topic/golang-nuts/apNcACpl_fI

antonmedv:
Hi, 

As I understand you are trying to call some like this expression? `foo(nil)` and it got panic? 

But uncertainly your PR breaks a lot of other functionality. I think we need to find another solution. 
As well as adding tests for this case. One of solution test in nil is poped and pass something typed, like `*int(nil)`, or type check argument on compile-time and pass correct type there.

```go 
package main

import (
	"fmt"
	"reflect"
)

func print(v interface{}) {
	fmt.Println(v)
}

func main() {
	var in interface{}
	f := reflect.ValueOf(print)

	in = "hello"
	print(in)
	f.Call([]reflect.Value{reflect.ValueOf(in)})

	in = 123
	print(in)
	f.Call([]reflect.Value{reflect.ValueOf(in)})

	var in2 *int = nil
	print(in2)
	f.Call([]reflect.Value{reflect.ValueOf(in2)})
}
```
https://play.golang.org/p/4MfgoNqymHS

wintermute-cds:
We use map[string]interface{} as enviroment where some optional fileds can be nil (not in map) or have different types (string, bool, flaot64 etc). Our funcs and overloaded operators use interface{} as argument type, thats why we need it. I have fixed the broken test by ensuring backwards compatibility (with a check for nil when we pop from stack). Is it OK for you?

wintermute-cds:
It passes all tests now, also our tests.

wintermute-cds:
Added test that illustrates our test case:
`func TestOverloadedEaqualWithNil(t *testing.T) {
	equals := func(a, b interface{}) interface{} {
		return a == b
	}
	env := map[string]interface{}{"equal": equals}

	p, err := expr.Compile("a == b",
		expr.Env(env),
		expr.Operator("==", "equal"),
		expr.AllowUndefinedVariables(),
	)
	if err != nil {
		t.Fatal(err)
	}
	
	out, err := expr.Run(p, env)
	if err != nil {
		t.Fatal(err)
	}
	if out != true {
		t.Fatal("expected true")
	}
}`

vm/vm.go
```diff
@@ -250,7 +250,12 @@ func (vm *VM) Run(program *Program, env interface{}) interface{} {
 			call := vm.constant().(Call)
 			in := make([]reflect.Value, call.Size)
 			for i := call.Size - 1; i >= 0; i-- {
-				in[i] = reflect.ValueOf(vm.pop())
+				popped := vm.pop()
+				if popped == nil {
+					in[i] = reflect.ValueOf(&popped).Elem()
```
antonmedv: IN this branch we already know what parameter is nil, but what kind of type should be pass to function?

wintermute-cds: Without the trick reflect.Call failes as reflect.ValueOf(nil) produces zero reflect.Value struct (that is as specified in the documentation). However, reflect.Call does not allow zero reflect.Value as parameter. Therfore we use the trick as follows:
```
var x interface{} = nil
v := reflect.ValueOf(*x).Elem()
```
The v reflect.Value is NOT a zero value and is therfore allowed in reflect.Call. We can still recover the untyped nil by using v.Interface().

The trick does not influence typed nil:
```
var y *int = nil
fmt.Println(reflect.ValueOf(y).Interface() == reflect.ValueOf(*y).Elem().Interface())
```

See also https://play.golang.org/p/Hz3mW6IrGli

wintermute-cds: There was still problem with typed nil, the latest commit fixes it.
https://play.golang.org/p/MyAztoZaqhx

wintermute-cds:
Do you see some other problems? Could you make a release with this patch any time soon?

vm/vm.go
antonmedv: What about doing this so: 
```go
				param := vm.pop()
				if param == nil {
					param = &param
				}
				in[i] = reflect.ValueOf(param)
```

antonmedv: BTW, this is `false`:
```go
	var a interface{} = nil
	var b interface{} = nil
	z := &b
	fmt.Printf("%T(%#v) %T(%#v) is %v\n", a, b, a, b, a == z)
	// <nil>(<nil>) <nil>(<nil>) is false 
```
https://play.golang.org/p/WYhz019HAku

wintermute-cds: The reflect.Value object in this case would no longer represent nil but a pointer to nil, it would not fix our problem:
https://play.golang.org/p/bcRDIrnz_rN

wintermute-cds: Pointer to nil is not nil so your observation is correct. Thats what the .Elem() is for in the code, we dereference pointer to nil, so we get nil. It should be equivalent:
```
var x interface{} = nil
v1 := reflect.ValueOf(&x).Elem()
v2 := reflect.ValueOf(x)
```
but is not, v1 and v2 are different thing. Thats why it is a GO-tcha and we have the trick to work around this.

antonmedv:
Yes, will do now. Also I've refactored it a little bit. Please check it.

antonmedv:
Please, go to #87

wintermute-cds:
What you did dit not work for us. I have updated the test to reflect this:
```
func TestOverloadedEaqualWithNil(t *testing.T) {
	equals := func(a, b interface{}) interface{} {
		if a != nil {
			panic(fmt.Errorf("a is not nil"))
		}
		if b != nil {
			panic(fmt.Errorf("a is not nil"))
		}
		return a == b
	}
	env := map[string]interface{}{"equal": equals}

	p, err := expr.Compile("a == nil",
		expr.Env(env),
		expr.Operator("==", "equal"),
		expr.AllowUndefinedVariables(),
	)
	if err != nil {
		t.Fatal(err)
	}
	
	out, err := expr.Run(p, env)
	if err != nil {
		t.Fatal(err)
	}
	if out != true {
		t.Fatal("expected true")
	}
}
```

PR #87, opened 2020-02-20
Allow passing nil to functions as nil interface{}

antonmedv:


antonmedv:
Released at https://github.com/antonmedv/expr/releases/tag/v1.4.4

PR #89, opened 2020-02-20
Updated the test with operator overloading

wintermute-cds:


wintermute-cds:
It is hard to explain. It is an edge case that happens to be important to us.

antonmedv:
I’ll release new release tomorrow.

wintermute-cds:
It is realy hard to explain. Strange thing in Go. Thank you for looking at it.

antonmedv:
Released v1.4.5 with a fix. 

Also, consider adding your project to [Who is using Expr](https://github.com/antonmedv/expr#who-is-using-expr) section.

PR #99, opened 2020-03-16
Add expr.Patch

antonmedv:


emicklei:
that is an elegant solution and would work for my case as demonstrated by your testcase.

Issue #113, opened 2020-05-05
float coercion for negative ints


emicklei:
Calling a function that expects a float but is passed a non-negative integer works as expected. However it fails when passing a negative integer.

```
package main

// float coercion for negative ints

import (
	"fmt"

	"github.com/antonmedv/expr"
)

func main() {
	env := map[string]interface{}{
		"cnv": func(f float64) interface{} { return f },
	}
	code := `cnv(-1)`
	program, err := expr.Compile(code, expr.Env(env))
	if err != nil {
		panic(err)
	}
	output, err := expr.Run(program, env)
	if err != nil {
		panic(err)
	}
	fmt.Println(output)
}
```
reports
```
panic: reflect: Call using int as type float64 (1:1)
 | cnv(-1)
 | ^
```

antonmedv:
I see. Looks like type assignment not working for negative ints. Can you make PR?

emicklei:
sure I can. happy to contribute to this great pkg

emicklei:
after adding a test that demonstrates the issue, I spent quite some time using the debugger how to package works and to find out where in the execution flow (parse,check,run), changes could be made. At checker.go#418 , I see the call to isIntegerOrArithmeticOperation that is suspicious. I think the fix should be replacing the IntegerNode with a FloatNode in the parse tree with converted data. 
Do you think this is the right direction?

antonmedv:
if check with isIntegerOrArithmeticOperation is correct place. Here are some thoughts: `foo(1)` as we parse `1` we do not know what type it is: float or int. Take foo signature and set the correct type to IntegrNode. Same for `foo(1+2)`: both `1` and `2` is ether int of float. Same for `foo(-1)`.

emicklei:
i made the changes in the folding.  see https://github.com/emicklei/expr/commit/0fce8f99baf9e6eee0b273c00305aea4631adc83 .
the tests added now PASS

antonmedv:
👏 awesome!
Can you make pr with it?

PR #114, opened 2020-05-09
Issue113

emicklei:
solves #113 by changing the optimizer to take into account the function argument type when folding.


antonmedv:
Merged fix #115

PR #115, opened 2020-05-10
Emicklei issue113

antonmedv:


ernest-ag5:
@antonmedv why was my entry removed?

Issue #123, opened 2020-05-17
[Bug?] Handle extra whitespace in "not in" syntax


qu-way:
i.e.
case 1:
```text
"A"  not in ["a", "b", "c"]
```
works and returns `true`

case 2:
```text
"A" not  in ["a", "b", "c"]
```
got panic:
```
panic: unexpected token Operator("not") (1:5)
 | "A" not  in ["a", "b", "c"]
 | ....^
```

do you guys think it's a bug and needs fix? thanks!

antonmedv:
It will be cool to fix it.

nmondal:
I am looking into it. Seems like "not in" is .. hard. It can be made "soft" by splitting it into "apply over" *in* operator.
@antonmedv  that diagnostics, is that right ?

antonmedv:
Yes, not in is ope operator. Possible to split it.

nmondal:
@antonmedv can you please look at it, if it solves it?

antonmedv:
It does, but the solution is really hacky. Can't merge it in this way.

nmondal:
Let me take another stab at it.

antonmedv:
Take a look at accept.

PR #144, opened 2020-10-22
This fixes issue #123

nmondal:
I am not too sure if that IS the right way of doing things, but looks right.

### DIFF
--- a/parser/lexer/lexer.go
+++ b/parser/lexer/lexer.go
@@ -2,6 +2,7 @@ package lexer
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 	"unicode/utf8"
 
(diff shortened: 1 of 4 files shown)


parser/lexer/lexer.go
```diff
@@ -77,7 +78,15 @@ func (l *lexer) emit(t Kind) {
...
 }
 
+// theoretically we can, actually have ' ' and '\t' and even '\n' too
+var replaceBlanks = regexp.MustCompile(` +`)
+
 func (l *lexer) emitValue(t Kind, value string) {
+	// some operator might have " " like 'not   in' look at the regex
+	// in that case, just replace all " +" with " "
+	if t == Operator && strings.Contains(value, "  ") {
+		value = replaceBlanks.ReplaceAllString(value, " ")
```
antonmedv, line 88: This is hacky solution.

parser/lexer/state.go
```diff
@@ -125,7 +125,7 @@ loop:
 }
 
 func not(l *lexer) stateFn {
-	if l.acceptWord(" in") {
```
antonmedv, line 128: Better to implement proper lexer pass with accept.

PR #157, opened 2021-01-05
support temporary override variables of env.

zhuah:
this pr provide a way for user to override/attach new variables to env instance.

### DIFF
--- a/vm/runtime.go
+++ b/vm/runtime.go
@@ -15,7 +15,22 @@ type Call struct {
 
 type Scope map[string]interface{}
 
+type OverrideEnv interface {
+	GetOverride(name string) (interface{}, bool)
+	Env() interface{}
+}
+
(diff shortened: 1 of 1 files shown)


antonmedv:
So what is use case? Why not just use map or extend struct?

zhuah:
I'm creating a template engine, which should support `if`,`else`,`elif`,`for`, while in a `for` block, user will introduce some new local variable such as `idx`,`val` or `key`,`val`.

i found no way to wrap current env with some temporary variables, what i can do is copy all variables/functions from old env to a new map and adds the new variables, and it's not easy to have same behavior with the `vm` package.

antonmedv:
Why not just override local vars in env? Set old value back after loop scope. Keep a stack of loops with var overrides.

zhuah:
the old env may be a structure, and local variable name is defined by user.

PR #160, opened 2021-02-02
Use tcell/v2

dundee:


antonmedv:
Not working with go1.13 unfortunately  =(

PR #166, opened 2021-02-19
Fix handling of space(s) in between operator

### LINKED ISSUE #123: [Bug?] Handle extra whitespace in "not in" syntax

shmsr:
Thir PR should fix the handling of extra space `(U+0020)` between multi-word operators (eg. `not in`). I had another hacky solution though where I initially thought to sanitize the input in `parser/parser.go` in the `Parse` function with something like:
```go
func sanitize(input string) string {
       return strings.Join(strings.Fields(strings.TrimSpace(input)), " ")
}
```
But again this is a bad idea.

Now the implementation is generic enough and should help to parse other multi-word operators as well. @antonmedv does it look good to you?

Also did some nit fixes:
- Remove the unused state from lexer struct
- Fix formatting
- Use the type Kind for Number, String, Operator, Bracket, EOF

Fixes #123

### DIFF
--- a/expr_test.go
+++ b/expr_test.go
@@ -629,6 +629,14 @@ func TestExpr(t *testing.T) {
 			true,
 		},
 		{
+			`Two not in 0..1`,
+			true,
+		},
+		{
+			`Two not    in 0..1`,
+			true,
+		},
+		{
 			`Int32 in [10, 20]`,
 			false,
 		},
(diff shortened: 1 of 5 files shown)


antonmedv:
Awesome! Thanks.

PR #171, opened 2021-03-10
strict types check to OpEqualInt and OpEqualString

tufitko:
custom types, based on a string, don't compare

example:
```
package main

import (
	"fmt"
	"github.com/antonmedv/expr"
)

func main() {
	type customString string
	
	type env struct{
		String1 customString
		String2 customString
	}
	
	program, err := expr.Compile("String1 == String2", expr.Env(env{}))
	fmt.Println(err)
	
	result, err := expr.Run(program, env{})
	fmt.Println(result, err)
}
```

when I try to run program, I got error:
```
interface conversion: interface {} is main.customString, not string (1:9)
 | String1 == String2
 | ........^
```

OpEqualString doesn't work with custom types based on string (OpEqualInt has same problem)

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -265,17 +265,17 @@ func (c *compiler) UnaryNode(node *ast.UnaryNode) {
 }
 
 func (c *compiler) BinaryNode(node *ast.BinaryNode) {
-	l := kind(node.Left)
-	r := kind(node.Right)
+	l, lIsBasicType := kind(node.Left)
+	r, rIsBasicType := kind(node.Right)
 
 	switch node.Operator {
 	case "==":
 		c.compile(node.Left)
 		c.compile(node.Right)
 
-		if l == r && l == reflect.Int {
+		if l == r && l == reflect.Int && lIsBasicType && rIsBasicType {
 			c.emit(OpEqualInt)
-		} else if l == r && l == reflect.String {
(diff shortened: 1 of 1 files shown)


compiler/compiler.go
```diff
@@ -654,10 +654,14 @@ func encode(i uint16) []byte {
...
 	t := node.Type()
 	if t == nil {
-		return reflect.Invalid
+		return reflect.Invalid, false
 	}
-	return t.Kind()
+
+	k := t.Kind()
+	name := t.Name()
+	return k, name != "" && name != k.String()
```
antonmedv, line 666: Is it the right way of checking if it’s custom type?

tufitko: not really,
for string name = 'string' (same for other basic types: int='int', int8='int8', etc...),
name is empty for maps, slices, arrays, functions, channels,
but for anonymous structs name also empty.

another option check `t.PkgPath() != ""`, we will get similar behavior.

antonmedv: Maybe we also need to check for anonymous structs than? Or google some canonical way of doing it)

tufitko: I googled and found nothing :( 
I will assume that the best option is type assertion. But we only have `reflect.Type`:
`_, ok := reflect.Zero(t).Interface().(string)` - looks poor

Anyway, I inverted logic. Now flag detects basic types. It works for all basic types without exception.

PR #172, opened 2021-03-21
Bitwise operators

munnik:
Added bitwise operators the the package

### DIFF
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -254,6 +254,10 @@ func (v *visitor) BinaryNode(node *ast.BinaryNode) reflect.Type {
 		if isInteger(l) && isInteger(r) {
 			return reflect.SliceOf(integerType)
 		}
+	case "<<", ">>", "&", "^", "|":
+		if isInteger(l) && isInteger(r) {
+			return integerType
+		}
 
 	default:
 		return v.error(node, "unknown operator (%v)", node.Operator)
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -170,6 +170,11 @@ func (v *visitor) UnaryNode(node *ast.UnaryNode) reflect.Type {
 			return t
 		}
 
+	case "~":
+		if isInteger(t) {
+			return integerType
+		}
+
 	default:
 		return v.error(node, "unknown operator (%v)", node.Operator)
 	}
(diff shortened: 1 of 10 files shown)


antonmedv:
Nice, but what is reason to add such? And how casting will work? Example, Int16 & uint32?

munnik:
Reason: I use expr mainly to parse Modbus data, typical expression I would like to use are:

`273.15 + (registers[0] << 16 | registers[1]) / 1000.0`
`(registers[0] & 8) == 8`

The first one is now solved with:

`273.15 + (registers[0] * 65536 + registers[1]) / 1000.0`

but I don't know an easy way to solve the second one. Besides that the first expressions are easier to understand.


Casting: don't really now how to solve that in a clean way. Now you can only use unsigned integer types. Technically it wouldn't matter to expect signed integers as well but I think it gives unexpected behaviors for the binary not operator.

antonmedv:
I see. Let me test your solution. Also lolike ci failed.

aaditya-panik:
Hi @munnik 

Adding this snippet,

```
case OpLeftShift:
	code("OpLeftShift")

case OpRightShift:
	code("OpRightShift")

case OpBitwiseAnd:
	code("OpBitwiseAnd")

case OpBitwiseXor:
	code("OpBitwiseXor")

case OpBitwiseOr:
	code("OpBitwiseOr")

case OpBitwiseNot:
	code("OpBitwiseNot")
```

 after,

https://github.com/antonmedv/expr/blob/2c1881a9909453f9f1047886d9be0b94e0ba5c48/vm/program.go#L145-L147

should pass the tests. It wasn't able to resolve the opcodes for disassembly of bitwise operators.

munnik:
Hi @aaditya-panik,

Thank you for the suggestion, it fixed the tests.

aaditya-panik:
Hi @munnik, @antonmedv 

I noticed that using numeric literals for bitwise operations seems to throw an error because they are parsed as `int` and the operations check for `uint` type. I have another branch which I had used to fix this (https://github.com/aaditya-panik/expr/commit/d24c08d94a400a11db1660531ab28fee3d4c1c23), let me know what you think. (Yet to add some tests)

antonmedv:
Is it still actial? Not sure we want to add bitwise ops here.

Issue #147, opened 2020-11-21
Support for conditional chaining


asaf:
I have deep objects which may sometimes have nil on the chain.

In those cases, without conditional nil operator it becomes very difficult to write expressions.

It could be great to support conditional chaining using question mark such:

https://play.golang.org/p/O70145vAj0-

Best,
Asaf.

antonmedv:
Done! Implement and later fixed)

PR #173, opened 2021-03-22
Add nil safe operator

zachbadgett:
Takes care of #147 

Only thing this doesn't work on is when the env struct is predefined before compiling due to `OpEqualString` and `OpEqualInt` optimizations doing type casts.

Not completely happy with the way it's implemented, but it was the easiest to do without changing much core code.

### DIFF
--- a/ast/node.go
+++ b/ast/node.go
@@ -48,7 +48,8 @@ type NilNode struct {
 
 type IdentifierNode struct {
 	base
-	Value string
+	Value   string
+	NilSafe bool
 }
 
 type IntegerNode struct {
@@ -100,6 +101,7 @@ type PropertyNode struct {
 	base
 	Node     Node
 	Property string
+	NilSafe  bool
 }
 
 type IndexNode struct {
@@ -120,6 +122,7 @@ type MethodNode struct {
 	Node      Node
 	Method    string
 	Arguments []Node
+	NilSafe   bool
 }
 
 type FunctionNode struct {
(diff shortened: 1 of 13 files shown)


zachbadgett:
Hmm just noticed this also doesn't short circuit the chain, so something like this: `foo?.missing.test() == '' && bar['missing'] == nil` would still fail. I'll see what I can do to fix that...

antonmedv:
Awesome work! 👏 I'll try to take a closes look at this before merging.

guillaume-merindol-cko:
Any news on this PR? :)

antonmedv:
Sorry for delay! @davidklassen can you also take a look? Thanks.

davidklassen:
LGTM

antonmedv:
Found an error. In this case error should be returned, but instead <nil> is returned.

```
package main

import (
	"fmt"

	"github.com/antonmedv/expr"
)

type S struct {
	S *S
	A string
}

func main() {
		"s": S{S: &S{S: &S{A: "OK"}}},
	}

	code := `s?.S.NOPE.A`

	if err != nil {
		panic(err)
	}

	if err != nil {
		panic(err)
	}

	fmt.Println(output)
}
```
Part with `S.NOPE` should throw error as not nil safe. 

@zachbadgett can you take a look? Thanks.

antonmedv:
The problem is with parser:
```diff
go run github.com/antonmedv/expr/cmd/exe -ast <<< 'a?.b.c'
PropertyNode{
	Node: PropertyNode{
		Node: IdentifierNode{
			Value: "a",
			NilSafe: true,
		},
		Property: "b",
-		NilSafe: true,
	},
	Property: "c",
-	NilSafe: true,
}
```

antonmedv:
One year later, I think this is not error any more)))

PR #183, opened 2021-06-14
Numpy style array functions and operators

regel:
Hey Anton,

I wrote this extension to use Numpy array math operations in expr.

Let me know your feedback

Sebastien


antonmedv:
This is really nice work, but unfortunately have to be declined. 

First, it's already way too of sync with my recent rewrite of Expr.

Second, I think it should not be enabled by default and be implemented as an extension. 

I will definitely reiterate on numpy if there is more interest in such.

PR #193, opened 2021-08-03
feat: codegen to fetch from env without reflect

tufitko:
Fetch methods generates for all struct/map/array/string named types (exception is map types with unnamed not basic key type like `map[struct{...}]int`)
To generate just call exprgen with pkg paths as arguments: `exprgen pkg1 pkg2...`
After call, file `*pkg_name*_exprgen.go` will be created

What do you think about it?


antonmedv:
Hi 👋 
Awesome stuff! Let’s write documentation on it and merge.

tufitko:
Added!

antonmedv:
Thanks.

exprgen/exprgen.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+
+	echo(`// Code generated by exprgen. DO NOT EDIT.`)
+	echo(``)
+	echo(`package ` + pkgName)
+	echo(``)
+	echo(`--imports`)
+	echo(``)
+
+	echoRaw(`func toInt(a interface{}) int {
```
antonmedv, line 105: What about exporting it from vm.ToInt?

PR #197, opened 2021-08-09
Add support for ConstantNode in checker

mhr3:
After implementing an ast.Visitor (and using expr.Patch) which replaced some nodes with a ConstantNode, I was getting panics from checker.Check, as it didn't expect a ConstantNode, but this simple patch fixes that.


antonmedv:
Awesome! Can you add test for it?

mhr3:
added

PR #222, opened 2022-02-01
chore: Enable dependabot for automatic dependency updates

blkperl:


antonmedv:
I checked repo settings and looks like it’s already enabled.

blkperl:
@antonmedv I don't see any PRs for dependabot - https://github.com/antonmedv/expr/pulls/app%2Fdependabot

antonmedv:
As all deps are for testing/debugging. And expr dont have any deps in production build we can simple not bother with it)

PR #230, opened 2022-03-23
Adding Wish to the customer of Expr

fzhang-wish:
We (wish.com) are actively using Expr for our new golang Customer support chatbot service and is having a great experience :) so would like to acknowledge this in the README


antonmedv:
Awesome! Thanks a lot!

PR #239, opened 2022-05-21
fix ConstExpr to handle custom functions returning error

brenank:
Fixes https://github.com/antonmedv/expr/issues/238

If the custom func returns a non-nil error, that same error is returned from `expr.Compile` function.  
This mirrors the existing behavior for this in `expr.Run`.

Related context: https://github.com/antonmedv/expr/pull/159#issuecomment-769761579


brenank:
cc @ffenix113

Issue #154, opened 2020-12-15
How to access pointer value?


daisy1754:
I have a usecase where I want to distinguish 0 and nil so I'm trying to pass `*int` to expr. Below is code I tried and it doesn't work as I had expected. How can get the value of pointer? I want to write expression like `*num==1` or `num.Value==1` but neither seems to work

```
package main

import (
	"fmt"
	"github.com/antonmedv/expr"
)

func main() {
        num := 1
	env := map[string]interface{}{
		"num": &num,
	}

	
	p1, _ := expr.Compile(`num!=nil`, expr.Env(env))
	o1, _ := expr.Run(p1, env)
	fmt.Println(o1) // true
	
	p2, _ := expr.Compile(`num==1`, expr.Env(env))
	o2, _ := expr.Run(p2, env)
	fmt.Println(o2) // false
	
	p3, _ := expr.Compile(`num.Value==1`, expr.Env(env))
	o3, _ := expr.Run(p3, env)
	fmt.Println(o3) // nil
}
```

antonmedv:
Compare it to nil. Expr language supports pointer as normal variables.

daisy1754:
Hi Anton - I want to know if it's nil, 0 or other value.

antonmedv:
Use comparison operations ==. 
Don’t getting that is your problem.

daisy1754:
Maybe my example was not clear. When num is pointer to integer value, I want to first check if num is nil, if not I want to check value of int referenced by the pointer (`*num == 1` in golang's code). 
My question was how can I get value of referenced variable (*num) in expr.

antonmedv:
Just like var: num

daisy1754:
Hmm, I'm not really sure if we're on the same page. Can you rewrite below golang program in expr?

```
func foo(num *int) string {
	if num == nil {
		return "nil"
	} else if *num == 0 {
		return "zero"
	} else if *num == 1 {
		return "one"
	}
	return "other"
}
```

antonmedv:
Now I got you. Looks like a bug in expr with dereferencing of *int. Will take a look into it.
Currently this is not working, but should:
```go
package main

import (
	"fmt"
	"github.com/antonmedv/expr"
)


type Env struct {
	Num *int
}

func main() {
	code := `Num == nil ? "is nil" : (Num == 0 ? "zero" : "not zero")`

	program, err := expr.Compile(code, expr.Env(Env{}))
	if err != nil {
		panic(err)
	}

[...]

PR #245, opened 2022-06-19
Issue #154: Follow pointers to basic types.

ebusto:
Not the most elegant approach, but this solves for issue #154 reasonably well.


antonmedv:
I finally fixed it) 

But is a little bit different way: using a separate opcode and using dereference only if needed. 

See 8d50db26f99113901f2a23f07f50fe2e12343c9d

Issue #227, opened 2022-03-09
Extensions or helper for `time.Time`


GunRyu:
handling time in query is kinda hard, and what about support for the missing `Date`?

antonmedv:
Done!

Issue #203, opened 2021-09-12
sugar or syntax for date or datetime?


loveyou127:
e.g. `5 min ago` in expr, or `2 days long`.

antonmedv:
I was thinking of adding something for it for sure.

antonmedv:
`CompareDate(..., ...)`

antonmedv:
Done!

Issue #73, opened 2019-10-16
Add date operations


elliots:
It would be great  to be able to use date operations like `transaction.Date > now+5d`

Perhaps using https://github.com/timberio/go-datemath ?

antonmedv:
Hi,

This is what I was thinking of adding to the language at first but decided to keep it simple for a while. But now, maybe it's time to add this functionality now. 

Currently, this can be done by operator overloading:
```go
	type Segment struct {
		Date time.Time
	}
	type Request struct {
		Segments []Segment
		Before   func(a, b time.Time) bool
		Date     func(s string) time.Time
	}

	code := `Date("2001-01-01") < Segments[0].Date`

	program, err := expr.Compile(code, expr.Env(&Request{}), expr.Operator("<", "Before"))
	if err != nil {
		fmt.Printf("%v", err)
		return
	}

	request := &Request{
		Segments: []Segment{
			{Date: time.Date(2019, 7, 1, 0, 0, 0, 0, time.UTC)},
		},
		Before: func(a, b time.Time) bool {
			return a.Before(b)
		},
		Date: func(s string) time.Time {
			date, err := time.Parse("2006-01-02", s)
			if err != nil {
				panic(err)
			}
			return date
		},
	}

	output, err := expr.Run(program, request)
	if err != nil {
		fmt.Printf("%v", err)
		return
	}

	fmt.Printf("%v", output)

	// Output: true
```

elliots:
You're right, I can get 90% of the way there with the operator overloading and that package.

```golang
import (
	"log"
	"testing"
	"time"

	"github.com/antonmedv/expr"
	"github.com/stretchr/testify/require"
	"github.com/timberio/go-datemath"
)

type DateTransaction struct {
	ReportDate time.Time
}

type Request struct {
	TX DateTransaction
}

func (*Request) Today() time.Time {
	return time.Now().Truncate(24 * time.Hour)
}

func (*Request) Date(d string) time.Time {
	t, err := time.Parse("20060102", d)
	if err != nil {
		panic(err)
	}
	return t
}

func (*Request) Add(t time.Time, amount string) time.Time {
	log.Printf("adding %s to %s", amount, t)
	expr, _ := datemath.Parse("now+" + amount)
	return expr.Time(datemath.WithNow(t))
}

func (*Request) Subtract(t time.Time, amount string) time.Time {
	log.Printf("subtracting %s from %s", amount, t)
	expr, _ := datemath.Parse("now-" + amount)
[...]

PR #256, opened 2022-09-23
Adds support for common operators for time.Time

panthershark:
This should resolve the following issues. 

#73 
#203
#227

Before starting this PR, I attempted to add support for time operators, but ended up with an impossible state b/c I cannot support `interface{} -OP- interface{}` without losing the built-in functionality of the operators. The only option I could see was to copy/paste/extend [/vm/helpers.go](https://github.com/antonmedv/expr/blob/master/vm/helpers.go). 

I decided to search the issues and found that others have run into this too. 

---- 
With this PR, the operators `==, >=, <=, <, >` are included by default allowing users to compare `time.Time` instances, returning  bool. 

The `+` operator adds a `time.Duration` to `time.Time`, returning `time.Time`
The `-` operator returns a duration representing the diff between two `time.Time` instances.

---- 
Please lmk if there are any concerns with the implementation. Please lmk if you see any gaps in the tests.  I'll be happy to fix it up. 

:)

### DIFF
--- a/checker/types.go
+++ b/checker/types.go
@@ -2,6 +2,7 @@ package checker
 
 import (
 	"reflect"
+	"time"
 
 	"github.com/antonmedv/expr/ast"
 )
@@ -15,6 +16,8 @@ var (
 	arrayType     = reflect.TypeOf([]interface{}{})
 	mapType       = reflect.TypeOf(map[string]interface{}{})
 	interfaceType = reflect.TypeOf(new(interface{})).Elem()
+	timeType      = reflect.TypeOf(time.Time{})
+	durationType  = reflect.TypeOf(time.Duration(0))
 )
 
 func typeWeight(t reflect.Type) int {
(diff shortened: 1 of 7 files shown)


antonmedv:
Awesome. Thanks. Will try to review on weekend.

panthershark:
Sweet thanks!  

I added tests to expr_test.go this morning and found I needed to update the checker for this to be fully functional.  I think it is pretty close. I'm happy to fix up anything that I missed.  

Thanks for making this library.  :)

checker/types.go
```diff
@@ -136,7 +136,7 @@ func isTime(t reflect.Type) bool {
 			return true
 		}
 	}
-	return false
+	return isInterface(t)
```
antonmedv, line 139: Can we add test for this part?

panthershark: Added to checker_test.go in 5ea8ef9.

panthershark:
bump. Are there any other areas that need attention? Is there anything blocking accepting this PR?

antonmedv (approved):

antonmedv:
Awesome work) really nice. Also will be cool to see some docs.

PR #274, opened 2022-11-05
Add Fetcher interface

antonmedv:
Example of `vm.Fetcher` implementation. It looks like it is much worse.

Benchstat: https://github.com/antonmedv/expr/actions/runs/3399892066/jobs/5654124890#step:8:21


vm/vm.go
```diff
@@ -116,6 +120,13 @@ func (vm *VM) Run(program *Program, env interface{}) (out interface{}, err error
 		case OpFetchEnvFast:
 			vm.push(env.(map[string]interface{})[program.Constants[arg].(string)])
 
+		case OpFetcher:
+			a := vm.pop()
+			vm.push(a.(Fetcher).Fetch(program.Constants[arg].(string)))
+
+		case OpFetcherEnv:
+			vm.push(env.(Fetcher).Fetch(program.Constants[arg].(string)))
```
antonmedv, line 128: The resulting program for the real_world bench test uses this opcode. And looks like proper usage of reflect can be faster.

PR #281, opened 2022-11-11
Add OpEnv opcode

antonmedv:


antonmedv:
Looks like it’s constantly slower for 5%.

antonmedv:
Looks like the current approach with separate opcodes to read directly from env is faster by ~3%.

PR #284, opened 2022-11-12
Add dummy opcodes

antonmedv:


antonmedv:
diff/bench is inconclusive.

PR #289, opened 2022-11-18
Improve testing, remove generated code and add duration operators

lefinal:
This PR should add a number of improvements:
1. Table tests are now run as separate tests. The reason for this is how common Go tests are structured as well as human-readability in case of multiple failing tests: Each failing test is reported and not only a single one. This also makes running tests in parallel easier. On my machine, tests still run in less than 0.09s.
2. Support for Go's `time.Duration` is now available. This includes many math operations and comparisons. A required change for this was dropping generated helper functions. This is justified by the fact that `time.Duration`-support requires custom logic. For example:
- When performing addition with `time.Time`, we need to use methods from `time.Time`.
- When combined with numeric types, we treat `time.Duration` similarly to `int64`.
- When performing division, e.g., `time.Hour / 6`, we expect `time.Duration` as result type.

The new approach treats all `uint`/`int`-types as `int`. Code generation was removed and replaced by custom logic for each operation.

For further information, see commit messages.


antonmedv:
Why current approach with generated helpers not working?

antonmedv:
Also test improvements are ok. Happy to merge them. 

But helpers refactoring looks lore complicated then before.

PS It's better to send one thing in PR, as it's easier and more likely get merged.

lefinal:
Thanks for the feedback. Honestly, it's one of my first contributions, so I am happy for any advice.

Regarding the issue with helpers, there are a few examples for `time.Duration` (expressions here in Go syntax):

- `4 * time.Hour / 2 * time.Hour` should return `2`.
- `4 * time.Hour / 2` should return `2 * time.Hour` (type `time.Duration`).
- `time.Now() + 1 * time.Hour` should add 1 hour to the current timestamp, so return type is `time.Time` and `Add()` from `time.Time` is used.

Helpers are generated by creating all permutations of numeric types (with exceptions) and adding custom cases to the large `select`.
Here, all permutations need to be considered as well.
Adding custom logic only for `time.Duration` and `time.Time` is easy.
But when combining with the other numeric types, it gets tricky.
An example:
```go
switch a := a.(type) {
// <<Start of generated part.>>
case uint8:
	switch b := b.(type) {
    case uint8:
		// ...
    case time.Duration:
		// ...
    }
case time.Duration:
	switch b := b.(type) {
    case uint8:
    // ...
    case time.Duration:
    // ...
    }
// <<End of generated part.>>

// Here come the latest additions for adding time.Time-support:
case time.Time:
	switch b := b.(type) {
    case time.Time:
		// ...
    case time.Duration:
		// Add duration to time.
    }
// Here would be the custom time.Duration-case which does not work
// because of code generation already including this case.
}
```
If we want to add our custom logic, we'd need to add a case to the `switch`-statement for `time.Duration`.
However, this is already added via code generation.
If we exclude it from generation, we need to find a way of adding cases to all other numeric cases which were already generated.
For me personally, adding exceptions to the code generator seemed to add lots of complexity as many edge-cases would be needed to consider.

The new approach of course moves part of the complexity to the operator functions.
However, understanding behavior for a single operator and having the option to add custom behavior instantly without affecting all other operators seemed easier than understanding the effects of changes to code generation, possibly affecting all generated operators in unintended ways.

lefinal:
If improvements in code coverage are needed, just tell me and I will add more test cases for the changes.

antonmedv:
In `4 * time.Hour / 2 * time.Hour` example, is Go code. We are working on Expr code) 

And I belive this time of arithmetic is not needed. Most of the time it will used like: `now() - created_at < duration("1h")`.

Happy to accept test improvements. =)

lefinal:
And yeah, I'll add more test cases :) Expect changes tomorrow or over the course of the next week.

antonmedv:
> Additionally from my understanding, if used with time.Time, all operators only work, if time.Time is on the left side and time.Duration on the right side. 

This is a good argument. Let me think this through.

lefinal:
Any updates on this? Latest push should provide high coverage of all added functionality.

antonmedv:
Let’s try to benchmark new solution little bit. I will push new changes today, can you please rebase on them and we rerun diff/bench.

vm/runtime/conversions.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+	return aDur, bDur, true
+}
+
+// typeName is a type representation, avoiding plain reflect-types.
+type typeName int
+
+// Supported type names.
+const (
+	typeUInt = iota
```
tw1nk, line 173: `typeUInt typeName = iota`
is a bit nicer as they then are typed

lefinal: Absolutely yes! I forgot setting the type. Thank you.

lefinal:
> Let’s try to benchmark new solution little bit. I will push new changes today, can you please rebase on them and we rerun diff/bench.

Still waiting for pushed changes :)

antonmedv:
Yeah, still didn’t get the time to finish the changes. I hope to finish my changes this weekend.

lefinal:
Any updates on this?

antonmedv:
Please split to two PRs and rebase.

antonmedv:
Please split the PR and rebsase.

PR #308, opened 2023-01-20
Dont check map key type if undef vars are allowed

krvladislav:


checker/checker.go
```diff
@@ -393,7 +393,7 @@ func (v *visitor) MemberNode(node *ast.MemberNode) (reflect.Type, info) {
 		return anyType, info{}
 
 	case reflect.Map:
-		if !prop.AssignableTo(base.Key()) {
+		if v.config.Strict && !prop.AssignableTo(base.Key()) {
```
antonmedv, line 396: Should the check be `!isAny(prop)`? As from the test example **Key** is **any**.

expr_test.go
```diff
@@ -434,6 +434,29 @@ func ExampleAllowUndefinedVariables_zero_value_functions() {
 	// Output: [foo bar]
 }
 
+func ExampleAllowUndefinedVariables_no_map_type_check() {
```
antonmedv, line 437: Can we make this one as a test and not an example?

antonmedv (changes requested):

antonmedv:
Fixed via 5366e576a3a08788c5d3463e95a489d86c9d79fa

PR #314, opened 2023-01-22
Add OpCallSuper opcode

antonmedv:


antonmedv:
Decided to go with expr.Function instead.

PR #327, opened 2023-01-31
docs: fix syntax error

rickyggg:
fix syntax error

old syntax will trigger panic:
`panic: unexpected token Identifier("len") (1:22)`


antonmedv:
Well, this syntax for new version in master.

PR #328, opened 2023-02-01
Adds simple bool condition optimization

andrewbenton:
This change introduces simple boolean condition optimization when dealing with boolean `and`, `or`, and `==` expressions to help trim unnecessary AST nodes.

### DIFF
--- a/compiler/compiler_test.go
+++ b/compiler/compiler_test.go
@@ -112,11 +112,12 @@ func TestCompile(t *testing.T) {
 		{
 			`-1`,
 			vm.Program{
-				Constants: []interface{}{-1},
+				Constants: []interface{}{1},
 				Bytecode: []vm.Opcode{
 					vm.OpPush,
+					vm.OpNegate,
 				},
-				Arguments: []int{0},
+				Arguments: []int{0, 0},
 			},
 		},
 		{
@@ -232,7 +233,7 @@ func TestCompile(t *testing.T) {
 	}
 
 	for _, test := range tests {
-		program, err := expr.Compile(test.input, expr.Env(Env{}))
+		program, err := expr.Compile(test.input, expr.Env(Env{}), expr.Optimize(false))
 		require.NoError(t, err, test.input)
 
 		assert.Equal(t, test.program.Disassemble(), program.Disassemble(), test.input)
--- a/optimizer/fold.go
+++ b/optimizer/fold.go
@@ -42,6 +42,10 @@ func (fold *fold) Visit(node *Node) {
 			if i, ok := n.Node.(*FloatNode); ok {
 				patchWithType(&FloatNode{Value: i.Value}, n.Node.Type())
 			}
+		case "!", "not":
+			if a := toBool(n.Node); a != nil {
+				patch(&BoolNode{Value: !a.Value})
+			}
 		}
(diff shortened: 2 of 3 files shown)


optimizer/fold.go
```diff
@@ -211,6 +215,50 @@ func (fold *fold) Visit(node *Node) {
...
+				patch(&BoolNode{Value: true})
+			}
+		case "==":
+			{
+				a := toInteger(n.Left)
+				b := toInteger(n.Right)
+				if a != nil && b != nil {
+					patch(&BoolNode{Value: a.Value == b.Value})
+				}
+			}
```
andrewbenton, line 247: I skipped adding float/float and int/float precomputation here given the potential issues with floating point comparison, but I could add it in if desired.

compiler/compiler_test.go
```diff
@@ -232,7 +233,7 @@ func TestCompile(t *testing.T) {
 	}
 
 	for _, test := range tests {
-		program, err := expr.Compile(test.input, expr.Env(Env{}))
+		program, err := expr.Compile(test.input, expr.Env(Env{}), expr.Optimize(false))
```
andrewbenton, line 236: Optimizations needed to be disabled in order for the string, int, and bool comparisons to still be checking the compiler's output given an unoptimized tree.  If optimizations are enabled, they each only have `vm.OpTrue` for the bytecode.

antonmedv:
Another think to optimize:
```
all(list, predicate_a) && all(list, predicate_b)
```
to 
```
all(list, predicate_a && predicate_b)
```

andrewbenton:
That could be interesting.  A quick investigation shows that'd look like:

```yaml
# from: `all(input, {# startsWith "scan"}) and all(input, {len(#) > 3})`

BinaryNode:
  Operator: "and"
  Left:
    BuiltinNode:
      Name: "all"
      Arguments:
        - IdentifierNode: { Value: "input" }
        - ClosureNode:
            Node:
              BinaryNode:
                Operator: "startsWith"
                  Left:
                    PointerNode: {}
                  Right:
                    StringNode: { Value: "scan" }
  Right:
    BuiltinNode:
      Name: "all"
      Arguments:
        - IdentifierNode: { Value: "input" }
        - ClosureNode:
            Node:
              BinaryNode:
                Operator: ">"
                  Left:
                    BuiltinNode:
                      Arguments:
                        - PointerNode: {}
                  Right:
                    IntegerNode: { Value: 3 }
```

and the output would be:
```yaml
BuiltinNode:
  Name: "all"
  Arguments:
    - IdentifierNode: { Value: "input" }
    - ClosureNode:
        Node:
          BinaryNode:
            Operator: "and"
            Left:
              BinaryNode:
                Operator: "startsWith"
                Left:
                  PointerNode: {}
                Right:
                  StringNode: { Value: "scan" }
            Right:
              BinaryNode:
                Operator: ">"
                Left:
                  BuiltinNode:
                    Arguments:
                      - PointerNode: {}
                Right:
                  IntegerNode: { Value: 3 }
```

Would you be okay with accepting the current work?  I can try to add the `and` of `all` and `or` of `any` optimizations later.

antonmedv:
Sounds good 👍🏻

PR #348, opened 2023-03-09
Add type check for boolean expect.

### LINKED ISSUE #347: expr.AsInt allows any, expr.AsBool does not allow any
Hey @antonmedv, great library, and appreciate your contribution to the Open Source world!

I've been trying to port my own `tabloid` app to an expression library like yours (I tried `cel-go`, but it's still too cumbersome to use) since the library I'm using seems abandonware. 

I tried to create a function called `ready` that would take in a field. If the field says `"ready"`, then it would return true, or false otherwise.

When I tried to plug it in and tell the compiler that the returned expression should be a boolean, I got an error because the return value of an `expr.Function` is an `interface{}`, not a boolean (even if the underlying interface value is a boolean).

Here's an example code. This code works, but it does not provide the help of `expr.AsBool()` (as in, it won't pre-check if the expression does return a boolean or not):

```go
	_, err := expr.Compile("ready(ready)", expr.Function("ready", func(params ...any) (any, error) {
		return params[0].(string) == "ready", nil
	}))

	if err != nil {
		log.Fatal(err.Error())
	}
```

This code, on the other hand, does not work (note the `expr.AsBool()` call). It fails with `expected bool, but got interface {}`:

```go
	_, err = expr.Compile("ready(ready)", expr.AsBool(), expr.Function("ready", func(params ...any) (any, error) {
		return params[0].(string) == "ready", nil
	}))

	if err != nil {
		log.Fatal(err.Error())
	}
```

Runnable Go playground: https://go.dev/play/p/kAqIUjaLJhX

I'm working around it by simply compiling without the `AsBool` call, and just checking the returned value once the expression executes, but I feel it could be improved by checking the interface value. I don't know enough about the library though to know whether this would have any side effect.

patrickdappollonio:
I tried this and it should fix #347, but let me know if you would want to see more tests added here.

### DIFF
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -36,6 +36,10 @@ func Check(tree *parser.Tree, config *conf.Config) (t reflect.Type, err error) {
 			if !isNumber(t) && !isAny(t) {
 				return nil, fmt.Errorf("expected %v, but got %v", v.config.Expect, t)
 			}
+		case reflect.Bool:
+			if !isBool(t) && !isAny(t) {
+				return nil, fmt.Errorf("expected %v, but got %v", v.config.Expect, t)
+			}
 		default:
 			if t != nil {
 				if t.Kind() == v.config.Expect {


antonmedv:
Closing. See in https://github.com/antonmedv/expr/issues/347#issuecomment-1461952794

PR #349, opened 2023-03-09
runtime.Fetch: add string type assertion check before trying to check if the value is a method

### LINKED ISSUE #346: `interface conversion: interface {} is int, not string` when trying to index slice type
Hello,

When trying to upgrade expr from 1.9 to the latest version in [crowdsec](https://github.com/crowdsecurity/crowdsec), some of our automated tests failed with the following error:
```
WARN[09-03-2023 00:17:41] runtime error : interface conversion: interface {} is int, not string (1:27)
 | "192.168.9.212" in results[0].Overflow.GetSources()
 | ..........................^
```
After some debugging, I managed to narrow down the issue and reproduce it with the following code:
```go
package main

import (
	"fmt"

	"github.com/antonmedv/expr"
)

type EventsSlice []Event

func (b EventsSlice) Len() int {
	return 42
}

type Event struct {
	Foo string
}

func main() {
		"results": EventsSlice{{"bar"}},
	}

	code := `results[0].Foo`

	if err != nil {
		fmt.Printf("error compiling expression: %s\n", err)
		return
	}

	fmt.Printf("%s", program.Disassemble())

	if err != nil {
		fmt.Printf("error running expression: %s\n", err)
		return
	}

	fmt.Println(output)
}
```
and gives this output:
```
0	OpLoadFast	0	results
1	OpPush	1	0
2	OpFetch
3	OpFetchField	2	{Foo [0]}
error running expression: interface conversion: interface {} is int, not string (1:8)
 | results[0].Foo
 | .......^
```

It seems  that if a slice type has a non-pointer receiver,  you cannot index it (if you change the `Len()` definition to be a pointer receiver, expr has no issue using the variable).

This issue is probably related to https://github.com/antonmedv/expr/issues/271, which you did not manage to reproduce at the time.

Go version: `go version go1.20.1 linux/amd64`

blotus:
Fixes #346.

I'm not familiar enough with expr internals to know if this fixes the issue or hides the underlying cause.

Feel free to close it if you have a better solution.

### DIFF
--- a/vm/runtime/runtime.go
+++ b/vm/runtime/runtime.go
@@ -18,9 +18,11 @@ func Fetch(from, i interface{}) interface{} {
 
 	// Methods can be defined on any type.
 	if v.NumMethod() > 0 {
-		method := v.MethodByName(i.(string))
-		if method.IsValid() {
-			return method.Interface()
+		if methodName, ok := i.(string); ok {
+			method := v.MethodByName(methodName)
+			if method.IsValid() {
+				return method.Interface()
+			}
 		}
 	}
 
--- a/expr_test.go
+++ b/expr_test.go
@@ -1613,6 +1613,32 @@ func TestIssue271(t *testing.T) {
 	require.Equal(t, 1.0, output)
 }
 
+type Issue346Array []Issue346Type
+
+type Issue346Type struct {
+	Bar string
+}
+
+func (i Issue346Array) Len() int {
+	return len(i)
+}
+
+func TestIssue346(t *testing.T) {
+	code := `Foo[0].Bar`
+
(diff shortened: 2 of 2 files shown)


vm/runtime/runtime.go
```diff
@@ -18,9 +18,11 @@ func Fetch(from, i interface{}) interface{} {
 
 	// Methods can be defined on any type.
 	if v.NumMethod() > 0 {
-		method := v.MethodByName(i.(string))
-		if method.IsValid() {
-			return method.Interface()
+		if methodName, ok := i.(string); ok {
```
antonmedv, line 21: Yes, this is the right solution 👍🏻

PR #355, opened 2023-04-17
Fixed a compilation error.

PranavPeshwe:
There is no member named timestamp in Message.
Added some formatting for better looks.
The output now looks like this:
```
[17 Apr 23 20:34 IST	Oh My God!
 17 Apr 23 20:34 IST	How you doin?
 17 Apr 23 20:34 IST	Could I be wearing any more clothes?
]
```

### DIFF
--- a/docs/Getting-Started.md
+++ b/docs/Getting-Started.md
@@ -82,7 +82,7 @@ type Message struct {
 }
 
 func main() {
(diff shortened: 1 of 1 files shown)


PranavPeshwe:
@antonmedv Please do take a look.

PR #356, opened 2023-04-18
remove duplicate codes

Schneizelw:


antonmedv:
Nice catch! I wonder if there are some tools to find it automatically? ChatGPT?

Schneizelw:
> Nice catch! I wonder if there are some tools to find it automatically? ChatGPT?

Expr has a complex context, and I have tried using chatGPT3.5 with inaccurate results.

antonmedv:
GPT-4 if asked explicitly about code duplicates finds precisely this. Pretty awesome.

Issue #293, opened 2022-12-01
Add builtin string functions 


dk-lockdown:
+ upper(str)
+ lower(str)
+ left(str)
+ right(str)
+ lpad(str)
+ substring(str)
+ reverse(str)
...

purplefox:
Can we make this optional? Existing applications might not want to expose these functions to their end users.
For example, we are using expr to provide expressions in a data streaming language. We would like full control over the set of functions exposed to our users.
Would be nice to have some kind of config flag "disableBuiltIns" to prevent these functions being exposed when they are not wanted.

wvell:
Making this optional it would interfere with the docgen.CreateDoc. Then the docgen package also need a method of something like:
`AddStringBuiltins(*Context) *Context`

I think that would be a bit weird?

antonmedv:
But why optional? What is use case for it?

wvell:
Yeah I would like to see it in stdlib also, i do not have a use case for hiding it. Would you like to see it implemented here: https://github.com/antonmedv/expr/blob/master/builtin/builtin.go ?

dk-lockdown:
> But why optional? What is use case for it?

we use it calculate sql function

antonmedv:
But why optional? Let’s make them available all the time.

antonmedv:
Added bunch of functions:
```go
	{`string(5)`, "5"},
	{`string(5.5)`, "5.5"},
	{`string("5.5")`, "5.5"},
	{`trim("  foo  ")`, "foo"},
	{`trim("__foo___", "_")`, "foo"},
	{`trimPrefix("prefix_foo", "prefix_")`, "foo"},
	{`trimSuffix("foo_suffix", "_suffix")`, "foo"},
	{`upper("foo")`, "FOO"},
	{`lower("FOO")`, "foo"},
	{`split("foo,bar,baz", ",")`, []string{"foo", "bar", "baz"}},
	{`splitN("foo,bar,baz", ",", 2)`, []string{"foo", "bar,baz"}},
	{`splitAfter("foo,bar,baz", ",")`, []string{"foo,", "bar,", "baz"}},
	{`splitAfterN("foo,bar,baz", ",", 2)`, []string{"foo,", "bar,baz"}},
	{`replace("foo,bar,baz", ",", ";")`, "foo;bar;baz"},
	{`replace("foo,bar,baz,goo", ",", ";", 2)`, "foo;bar;baz,goo"},
[...]

PR #357, opened 2023-04-20
String builtins

andygello555:
Added the following builtins for string manipulation (as per #293 + comments):

- `upper(str)`: uses the `strings.ToUpper` method from the standard strings library
- `lower(str)`: uses the `strings.ToLower` method from the standard strings library
- `left(str, length)`: uses `runtime.Slice` to slice a string from the left up until the given length
- `right(str, length)`: uses `runtime.Slice` to slice a string from the right up until the given length
- `lpad(str, padLength, padString)`: uses `runtime.pad` to pad the left of a string with the given padding for the required number of characters
- `rpad(str, padLength, padString)`: uses `runtime.pad` to pad the right of a string with the given padding for the required number of characters
- `pad(str, padLength, padString)`: uses `runtime.pad` to pad both sides of a string with the given padding for the required number of characters. I.e. centering the string within the given padding.
- `substr(str, from, to)`: uses `runtime.Slice` to retrieve the substring within the given string that starts at index `from` and finishes at `to`.
- `reverse(str)`: reverses the given string
- `split(str, sep)`: splits the given string at the given separator using `strings.Split`

**All the methods above work on strings containing UTF-8 characters**

## Changes made to `builtin` package

- Opcodes for new builtins mentioned above
- Added new builtins to `builtin.Builtins` lookup
  - Each `builtin.Function` has a `Validate` implementation
- `validateArg` + `validateArgs` methods which validate multiple args according to a set of `reflect.Kind`s for each argument
- Added tests for new builtins

## Changes made to `vm` package

- Handle new builtin opcodes in `OpBuiltin` switch
- This uses the new functions in `runtime` described below

## Changes made to `runtime` package

- Updated `runtime.Slice` to support slicing of UTF-8 strings
- Added `runtime.Upper`, which calls `strings.ToUpper`, for new `upper` builtin
- Added `runtime.Lower`, which calls `strings.ToLower`, for new `lower` builtin
- Added `runtime.Left` for new `left` builtin
- Added `runtime.Right` for new `right` builtin
- Added the `runtime.pad` function which is called by the `runtime.LPad`, `runtime.RPad`, and `runtime.Pad` implementations
- Added `runtime.LPad` for new `lpad` builtin
- Added `runtime.RPad` for new `rpad` builtin
- Added `runtime.Pad` for new `pad` builtin
- Added `runtime.Substring` for new `substr` builtin
- Added `runtime.Reverse` for new `reverse` builtin
- Added `runtime.Split` for new `split` builtin
- Added the `runtime.checkString` helper to assert check if `interface{}` == `string` and panic if not

## Docs

Have not yet added new builtins to docs


antonmedv:
Done. I've added most of the string builtins.

PR #364, opened 2023-04-26
add @ in lexer

tooptoop4:
needed to accept stringified json: https://github.com/tidwall/gjson/issues/260#issuecomment-1012294762

fixes https://github.com/argoproj/argo-workflows/issues/8734

### DIFF
--- a/parser/lexer/state.go
+++ b/parser/lexer/state.go
@@ -21,6 +21,8 @@ func root(l *lexer) stateFn {
 			l.error("%v", err)
 		}
 		l.emitValue(String, str)
+	case r == '@':
+		l.emitValue(String, "@")
 	case '0' <= r && r <= '9':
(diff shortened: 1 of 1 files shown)


jingyugao:
hi,what's the process of this pr. I met this problem too. If you are not busy, I can add some tests for this pr.

tooptoop4:
@jingyugao please take over this

antonmedv:
The `@` char is considered incorrect in Expr. Same in most programming languages. 

Here is a few solution you can use in Expr:

- Use `[` `]` brackets
  ```js
  foo.bar["@baz"]
  ```
- Use `get()` function
  ```js
  get(foo.bar, "@baz")
  ```
- To get top-level var with `@` use `$env` variable.
  ```js
  $env["@foo"].bar
  ```

PR #367, opened 2023-05-05
Add support for multiline string literals

spaceweasel:
Currently, an env variable can hold a multiline string, for example:
```go
func main() {
		"lines": `Here are
two lines`,
	}

	out, _ := expr.Eval("lines", env)
	fmt.Print(out)
}
```
Which prints out:
```
Here are
two lines
```
Whereas if I want to use a multiline string _literal_, then an error is returned.
```go
	out, err := expr.Eval("'Here are\ntwo lines'", env)
	if err != nil {
		panic(err)
	}
```
```
panic: literal not terminated (2:1)
 | two lines'
 | ^
```
A workaround is to use double backslashes:
```go
	out, err := expr.Eval("'Here are\\ntwo lines'", env)
```
Whilst this gives the desired output in this situation, it is impractical for our case.
Apart from the excessive line length, there are other issues like readability.

We have a requirement where the input strings do not exist in Go code, but are read from other sources (e.g. yaml files) and use the `map` function on a collection. An example would be something like this, but with many more lines and arguments:
```
map(things, sprintf(`
Thing Name: %s
Thing Size: %d`, 
.name, 
.size))
```
This PR adds support for multiline string literals through the backtick character, so it will handle our situation as well as more simple ones:
```go
	out, err := expr.Eval("`Here are\ntwo lines`", env)
```

### DIFF
--- a/parser/parser_test.go
+++ b/parser/parser_test.go
@@ -24,6 +24,10 @@ func TestParse(t *testing.T) {
 			&StringNode{Value: "str"},
 		},
 		{
+			"`multi\nline\nstring`",
+			&StringNode{Value: "multi\nline\nstring"},
+		},
+		{
 			"3",
 			&IntegerNode{Value: 3},
 		},
--- a/parser/lexer/lexer.go
+++ b/parser/lexer/lexer.go
@@ -206,7 +206,7 @@ func (l *lexer) scanEscape(quote rune) rune {
 func (l *lexer) scanString(quote rune) (n int) {
 	ch := l.next() // read character after quote
 	for ch != quote {
-		if ch == '\n' || ch == eof {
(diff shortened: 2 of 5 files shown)


antonmedv:
I'd like to use `` ` `` char for  raw strings. Now only multiline strings. Let's implement those instead.

parser/parser_test.go
```diff
@@ -23,6 +23,10 @@ func TestParse(t *testing.T) {
 			`"str"`,
 			&StringNode{Value: "str"},
 		},
+		{
+			"`multi\nline\nstring`",
```
antonmedv, line 27: Tests for raw strings.

antonmedv (changes requested):

antonmedv:
Implemented in #485

PR #373, opened 2023-05-19
Implemented env[] keyword and tests.

mdmcconnell:
For consideration:  The env[] keyword allows an expression to access identifiers in env directly, with any name desired, including numbers, spaces, special characters, etc.  Syntax env['string identifier'] and env["string identifier"] are allowable.

### DIFF
--- a/parser/lexer/state.go
+++ b/parser/lexer/state.go
@@ -155,15 +155,7 @@ func literalIdentifier(l *lexer) stateFn {
 				return l.error("env keyword with no closing bracket")
 			}
 		} else {
-			for r := l.next(); r != ']' && r != eof; r = l.next() {
-			}
-			if r == eof {
-				return l.error("env keyword with no closing bracket")
-			} else {
-				l.backup()
-				l.emit(Identifier)
-				l.next()
-			}
+			return l.error("env keyword must have string index")
 		}
 	} else {
 		l.backup()
--- a/parser/lexer/state.go
+++ b/parser/lexer/state.go
@@ -6,11 +6,6 @@ import (
 
 type stateFn func(*lexer) stateFn
 
-// The lexer works as a state machine, creating and emitting tokens
-// as it proceeds through the input file (string).  State changes with the
-// context of what characters have been encountered so far. This is the
-// initial state; when a change is detected, it returns the function  for the
-// relevant next state to proceed.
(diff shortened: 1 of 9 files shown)


antonmedv:
Will review this.

parser/lexer/state.go
```diff
@@ -128,6 +135,43 @@ loop:
 	return root
 }
 
+// Process what comes after the env keyword, which must be any
+// set of characters or a string literal, enclosed in square brackets
+func literalIdentifier(l *lexer) stateFn {
```
antonmedv, line 140: Let's not introduce new syntax. We can just require to use strings:
```
env['foo bar']
```

antonmedv (changes requested):

parser/lexer/state.go
mdmcconnell: Fair enough - it's not worthwhile to have new syntax for the sake of a two character savings.

mdmcconnell:
Bump - this is working well, if you'd like to pull the code.

parser/lexer/state.go
```diff
@@ -119,6 +124,8 @@ loop:
 				return not
 			case "in", "or", "and", "matches", "contains", "startsWith", "endsWith":
 				l.emit(Operator)
+			case "env":
+				return literalIdentifier
```
antonmedv, line 128: Let’s implement this without altering lexer.

antonmedv (changes requested):

parser/lexer/state.go
mdmcconnell: I feel that handling tokens within lexer is a clear, simple way to accomplish this.  Is the concern that too much interpretation of "meaning" of expressions moves into lexer?  Other keywords in, matches, contrains, startWith, endsWith are also interpreted within lexer.  I will look at an alternative.  Perhaps env can be a normal indexed identifier that parser will autmatically point at "env".  I suspect that will cause copmlications worrying about the type of the underlying value that the index returns; we'll see.

antonmedv: env is just a variable.

mdmcconnell:
Have moved env from lexer to parser.

mdmcconnell:
Added about env note in Language Definition.

PR #381, opened 2023-06-08
feat: Add context.Context support for methods and functions

thomascoquet:
Idea is to support functions of the form `func(context.Context, any...) (any, error)` (c.f. https://github.com/antonmedv/expr/issues/380).

Changes are lightweight: 

- A new OpCode is introduced for `builtin.Function` having `context.Context` as first argument,
- This setting can be set using `&builtin.Function{Context: true}`,
- For methods, type check was modified to account for such kinds of method and the runtime was updated accordingly.

The original API is not modified.

The use cases include:

- doing interruptable work with `expr`  (requests, etc.),
- my personal use case is keeping track of created objects, using the context as a poor man's DI.

Alternatives using methods:
```go
type env struct{
    context.Context
}

func (e *env) Do(s string) string {
    return s
}
```
lacks the type flexibility provided by `builtin.Function`.

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -12,9 +12,13 @@ var (
 )
 
 type Function struct {
-	Name     string
-	Func     func(args ...interface{}) (interface{}, error)
-	Opcode   int
+	Name    string
+	Func    func(args ...interface{}) (interface{}, error)
+	Opcode  int
+	Context bool
+	// If Context is true, the reflect.Type of context.Context shall not be included
+	// in the signatures in Function.Types.
+	// Similarly, context.Context will not be included when calling Function.Validate.
 	Types    []reflect.Type
 	Validate func(args []reflect.Type) (reflect.Type, error)
 }
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -1,6 +1,7 @@
 package builtin
 
 import (
+	"context"
 	"fmt"
 	"reflect"
 )
@@ -12,15 +13,12 @@ var (
 )
 
 type Function struct {
-	Name    string
-	Func    func(args ...interface{}) (interface{}, error)
-	Opcode  int
-	Context bool
-	// If Context is true, the reflect.Type of context.Context shall not be included
-	// in the signatures in Function.Types.
-	// Similarly, context.Context will not be included when calling Function.Validate.
-	Types    []reflect.Type
-	Validate func(args []reflect.Type) (reflect.Type, error)
+	Name            string
+	Func            func(args ...interface{}) (interface{}, error)
+	FuncWithContext func(ctx context.Context, args ...interface{}) (interface{}, error)
+	Opcode          int
+	Types           []reflect.Type
+	Validate        func(args []reflect.Type) (reflect.Type, error)
 }
 
 const (
--- a/docs/Getting-Started.md
+++ b/docs/Getting-Started.md
@@ -174,3 +174,18 @@ Here is another example with a few function signatures:
 		new(func(string) int),
 	)
 ```
+
+Functions taking `context.Context` as first argument are also accepted:
+
+```go
+	username := expr.FunctionWithContext(
+		"username",
+		func(params ...any) (any, error) {
+			ctx := params[0].(context.Context)
+			return ctx.Value("user"), nil
+		},
+		new(func() string),
+	)
+```
+
+Program shall be run using `expr.RunWithContext`.
--- a/docs/Getting-Started.md
+++ b/docs/Getting-Started.md
@@ -180,11 +180,10 @@ Functions taking `context.Context` as first argument are also accepted:
 ```go
 	username := expr.FunctionWithContext(
 		"username",
-		func(params ...any) (any, error) {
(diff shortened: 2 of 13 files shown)


builtin/builtin.go
```diff
@@ -12,9 +12,13 @@ var (
 )
 
 type Function struct {
-	Name     string
-	Func     func(args ...interface{}) (interface{}, error)
-	Opcode   int
+	Name    string
+	Func    func(args ...interface{}) (interface{}, error)
+	Opcode  int
+	Context bool
```
antonmedv, line 18: Let's add a new field, instead of this `Context` field. 

    FuncWithContext func(ctx, args ...interface{}) (interface{}, error)

docs/Getting-Started.md
```diff
@@ -174,3 +174,18 @@ Here is another example with a few function signatures:
 		new(func(string) int),
 	)
 ```
+
+Functions taking `context.Context` as first argument are also accepted:
+
+```go
+	username := expr.FunctionWithContext(
+		"username",
+		func(params ...any) (any, error) {
```
antonmedv, line 183: Same here, the expr.FunctionWithContext can be with signature of

    func (ctx, params...) any, error

expr_test.go
```diff
@@ -1736,6 +1741,28 @@ func TestFunction(t *testing.T) {
 	assert.Equal(t, 20, out)
 }
 
+func TestFunctionWithContext(t *testing.T) {
+	add := expr.FunctionWithContext(
+		"add",
+		func(p ...interface{}) (interface{}, error) {
+			out := 0
+			fmt.Println("p", len(p))
+			for _, each := range p[1:] {
```
antonmedv, line 1750: Let's also add the test of what actually uses the context feature. Something with terminating.

vm/vm.go
```diff
@@ -14,20 +15,27 @@ import (
...
+	return RunWithContext(context.Background(), program, env)
+}
+
+func RunWithContext(ctx context.Context, program *Program, env interface{}) (interface{}, error) {
 	if program == nil {
 		return nil, fmt.Errorf("program is nil")
 	}
-
-	vm := VM{}
+	vm := VM{ctx: ctx}
```
antonmedv, line 33: Currently, vm doesn't do anything with context. Only passed to some functions. 

But **expr.RunWithContext** sounds like VM should be stopped (for example on deadline).

vm/vm.go
```diff
@@ -297,7 +305,19 @@ func (vm *VM) Run(program *Program, env interface{}) (_ interface{}, err error)
 		case OpCall:
 			fn := reflect.ValueOf(vm.pop())
 			size := arg
-			in := make([]reflect.Value, size)
+			fnt := fn.Type()
+			var (
+				args []reflect.Value
+				in   []reflect.Value
+			)
+			if fnt.NumIn() > 0 && fn.Type().In(0) == contextType {
```
antonmedv, line 313: Let's not introduce context passing for general OpCall. 

Let's allow context to be passed only to funcs defined via `expr.FunctionWithContext` in OpCallContext.

antonmedv (changes requested):

antonmedv:
Overall a good PR. I have a few comments.

thomascoquet:
Hello Anton,
I applied the requested changes. I chose to use a wrapper to use the same func slice in `*vm.Program` instead of duplicating it.
Thomas

antonmedv:
Now ai think why do we need to pass context to functions via expr.VM. 

This passing of context complicates vm arguments handling. 

The VM itself doesn’t need context as I understand right now. 

Maybe there’s better methods of passing context to user defined functions?

antonmedv:
Or we can implement this via our own custom patcher. I think this is the right solution.

antonmedv:
### Added a doc on how to add context via patcher. 

- https://github.com/antonmedv/expr/blob/master/docs/Visitor-and-Patch.md#patch

antonmedv:
FYI. I've implemented my own version of context patcher and added a config option https://github.com/antonmedv/expr/blob/f92bb7949de064a05d5b3a61d1b07abef0c521bb/expr.go#L157-L161

```go
func ExampleWithContext() {
	env := map[string]any{
		"fn": func(ctx context.Context, _, _ int) int {
			// An infinite loop that can be canceled by context.
			for {
				select {
				case <-ctx.Done():
					return 42
				}
			}
		},
		"ctx": context.TODO(), // Context should be passed as a variable.
	}

	program, err := expr.Compile(`fn(1, 2)`,
		expr.Env(env),
		expr.WithContext("ctx"), // Pass context variable name.
	)
	if err != nil {
		fmt.Printf("%v", err)
		return
	}

	// Cancel context after 100 milliseconds.
	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond*100)
	defer cancel()

	// After program is compiled, context can be passed to Run.
	env["ctx"] = ctx

	// Run will return 42 after 100 milliseconds.
	if err != nil {
		fmt.Printf("%v", err)
		return
	}

	fmt.Printf("%v", output)
	// Output: 42
}
```

PR #382, opened 2023-06-10
env[] keyword implemented in parser

mdmcconnell:
The env[] keyword allows an expression to access identifiers in env directly, with any name desired, including numbers, spaces, special characters, etc. Syntax env['string identifier'] and env["string identifier"] are allowable.

### DIFF
--- a/expr_test.go
+++ b/expr_test.go
@@ -1811,6 +1811,82 @@ func TestEval_nil_in_maps(t *testing.T) {
 	})
 }
 
+// Test the use of env keyword.  Forms env[] and env[”] are valid.
+// The enclosed identifier must be in the expression env.
+func TestEnv_keyword(t *testing.T) {
+	env := map[string]interface{}{
+		"space test":             "ok",
+		"space_test":             "not ok", // Seems to be some underscore substituting happening, check that.
+		"Section 1-2a":           "ok",
+		"2015 Information Table": "ok",
+		"%*worst function name ever!!": func() string {
+			return "ok"
+		}(),
+		"env": func(string) string {
+			return "env func ok"
+		},
+		"1":   "o",
+		"2":   "k",
+		"num": 10,
+	}
+
+	// No error cases
+	var tests = []struct {
+		code string
+		want interface{}
+	}{
+		{"env['space test']", "ok"},
+		{"env[space test]", "ok"},
+		{"env[Section 1-2a]", "ok"},
+		{`env['2015 Information Table']`, "ok"},
+		{"env[%*worst function name ever!!]", "ok"},
+		{"env('Confusing!')", "env func ok"}, // not keyword, but some function named env
+		{"env[1] + env[2]", "ok"},
+		{"1 + env[num] + env[num]", 21},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.code, func(t *testing.T) {
+
+			program, err := expr.Compile(tt.code, expr.Env(env))
+			require.NoError(t, err, "compile error")
+
+			got, err := expr.Run(program, env)
+			require.NoError(t, err, "execution error")
+
+			assert.Equal(t, tt.want, got, tt.code)
+		})
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.code, func(t *testing.T) {
+			got, err := expr.Eval(tt.code, env)
+			require.NoError(t, err, "eval error: "+tt.code)
+
+			assert.Equal(t, tt.want, got, "eval: "+tt.code)
+		})
+	}
+
+	// error cases
+	tests = []struct {
+		code string
+		want interface{}
+	}{
+		{"env[this is a problem", "bad"},
+		{"env['also a problem'", "bad"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.code, func(t *testing.T) {
+
+			_, err := expr.Eval(tt.code, expr.Env(env))
+			require.Error(t, err, "compile error")
+
+		})
+	}
+
+}
+
 type Bar interface {
 	Bar() int
 }
--- a/expr_test.go
+++ b/expr_test.go
@@ -1815,10 +1815,10 @@ func TestEval_nil_in_maps(t *testing.T) {
 // The enclosed identifier must be in the expression env.
 func TestEnv_keyword(t *testing.T) {
 	env := map[string]interface{}{
-		"space test":             "ok",
-		"space_test":             "not ok", // Seems to be some underscore substituting happening, check that.
-		"Section 1-2a":           "ok",
-		"2015 Information Table": "ok",
+		"space test":                       "ok",
+		"space_test":                       "not ok", // Seems to be some underscore substituting happening, check that.
+		"Section 1-2a":                     "ok",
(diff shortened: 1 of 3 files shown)


antonmedv:
What about this case?

```
env[anotherVar]
```

mdmcconnell:
I think you are asking "what if the env[] index is an identifier, not a string?".  Previously, we had dicussed along the lines of "env[] index must be a string", but allowing an identifier that evaluates to a string seems reasonable, if more complex.  I will look at that.

antonmedv:
Also what about env[“foo” + “bar”]?

mdmcconnell:
That is another case of "something that evaluates to a string", versus "a string" which I had intended.  env[strfun(stuff)] would be another example, etc.  It's not the original idea, but I think a reasonable expectation.  I will have to implement further "downstream" in expr execution.

mdmcconnell:
Back to this - have now implemented so that any index that evaluates to a string.  I had to add a new opcode, since evaluation of the index may not happen until running the expression.   This will not be backwarsd compatible with expressions that have a an array or map named "env" in their environment; I think that is common with expr keywords - I can not for example define a function named "len".

checker/checker.go
```diff
@@ -132,7 +132,7 @@ func (v *visitor) IdentifierNode(node *ast.IdentifierNode) (reflect.Type, info)
 	if fn, ok := v.config.Functions[node.Value]; ok {
 		// Return anyType instead of func type as we don't know the arguments yet.
 		// The func type can be one of the fn.Types. The type will be resolved
-		// when the arguments are known in CallNode.
+		// when the arguments areknown in CallNode.
```
antonmedv, line 135: Typo.

mdmcconnell: Doh!  Sorry.

checker/checker.go
```diff
@@ -429,7 +432,9 @@ func (v *visitor) MemberNode(node *ast.MemberNode) (reflect.Type, info) {
 		return t, info{}
 
 	case reflect.Array, reflect.Slice:
-		if !isInteger(prop) && !isAny(prop) {
+		if isInteger(prop) || isAny(prop) {
```
antonmedv, line 435: Let's keep original if and do not flip condition.

compiler/compiler.go
```diff
@@ -491,19 +491,25 @@ func (c *compiler) MemberNode(node *ast.MemberNode) {
...
 
-	c.compile(base)
-	if node.Optional {
-		ph := c.emit(OpJumpIfNil, placeholder)
-		c.chains[len(c.chains)-1] = append(c.chains[len(c.chains)-1], ph)
-	}
-
-	if op == OpFetch {
+	// Implementation of direct env access
+	if ident, ok := base.(*ast.IdentifierNode); ok && ident.Value == "env" {
```
antonmedv, line 495: What about implementing this in the IdentifierNode directly? Let's add the opcode `OpLoadEnv` which will put the env on the stack and then use standard MemberNode implementation. 

Right now next expr features will not work:

```
env.FooBar
env?.optionalValue
```

expr_test.go
```diff
@@ -1811,6 +1811,97 @@ func TestEval_nil_in_maps(t *testing.T) {
...
+	// No error cases
+	var tests = []struct {
+		code string
+		want interface{}
+	}{
+		{"env['space test']", "ok"},
+		{"env['Section 1-2a']", "ok"},
+		{`env["c:\\ndrive\\2015 Information Table"]`, "ok"},
+		{"env['%*worst function name ever!!']", "ok"},
+		{"env('Confusing!')", "num"}, // not keyword, but some function named env
```
antonmedv, line 1852: This is a confusing part. Let's just forbid using the `env` keyword for anything else. 

We also can check this forbid rule on Compile step.

expr_test.go
```diff
@@ -1811,6 +1811,97 @@ func TestEval_nil_in_maps(t *testing.T) {
...
+
+	// error cases
+	tests = []struct {
+		code string
+		want interface{}
+	}{
+		{"env[this is a problem", "bad"},
+		{"env['also a problem'", "bad"},
+		{"env[space test]", "bad"},
+		{"env[1] + env[2]", "bad"},
```
antonmedv, line 1891: Those env tests are obsolete. They are already covered by our Lexer tests. No not need to test it here one more time.

antonmedv (changes requested):

antonmedv:
Overall a good PR. Much simple implementation. 👍🏻

expr_test.go
mdmcconnell: I did this to preserve some backwards compatibility with variables called env, though it will be less confusing to disallow that case.  Note: other keywords may be used in this way - I can define a string called "len" or "filter" without error.

checker/checker.go
```diff
@@ -383,6 +383,9 @@ func (v *visitor) ChainNode(node *ast.ChainNode) (reflect.Type, info) {
 }
 
 func (v *visitor) MemberNode(node *ast.MemberNode) (reflect.Type, info) {
+	if an, ok := node.Node.(*ast.IdentifierNode); ok && an.Value == "env" {
+		return anyType, info{}
```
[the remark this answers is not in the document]
mdmcconnell: Yes, makes sense.

checker/checker.go
mdmcconnell: Ok.

expr_test.go
mdmcconnell: Ok.

expr_test.go
antonmedv: > I can define a string called "len" or "filter" without error.

This is a thing to fix in the future. Disallow by default and only allow with `expr.AllowOverrideStandard()` option or something like this.

mdmcconnell: A convenient place to check for keyword misuse is in types_table.go, when all map keys and struct tags are examined.

antonmedv:
Nice work! 🎉

antonmedv:
Completely forgot. Please, write documentation for this feature. Thanks!

mdmcconnell:
Yes, of course - will write some documentation hopefully before next week.  Thank you for your patience while I've been learning how expr package works!

antonmedv:
Also: will be nice to add this to expr/docgen

harkli:
> The env[] keyword allows an expression to access identifiers in env directly, with any name desired, including numbers, spaces, special characters, etc. Syntax env['string identifier'] and env["string identifier"] are allowable, as is any index that evaluates to a string.

env["env"] will cause panic:  attempt to misuse env keyword as env map key

func TestEnv(t *testing.T) {
		"env": "demo",// panic!!!
	}
	p, e := expr.Compile(`env["env"] == "demo"`, expr.Env(env))
	o, e := expr.Run(p, env)
	fmt.Println(o)
	fmt.Println(e)
}

mdmcconnell:
That is WAD - it's not allowed to call any variable "env" (or other identifier) because the behaviour becomes indeterminate.  Which should have precedence, the keyword or the user-defined variable?  Should env[] recursively map into itself?  There is some discussion above around 9-Jul.  At the moment, other keywords like "len" can be overloaded, and there is intention to disallow that as well.

PR #386, opened 2023-06-29
feature: support raw string with char `

zhuliquan:
support raw string with char '` ' like golang, which it's helpful for regex matching


antonmedv:
Awesome! Will review the PR.

PR #452, opened 2023-10-04
docs: categorize Language Definition functions

agilgur5:
## Motivation

Since 1.14 and 1.15 added a bunch of new operators / functions, I thought that categorizing them would help discoverability
- in particular, some of the older functions were roughly categorized (all array functions in a row for instance), but newer ones seemed to have just been added to the end of the whole list

## Modifications

- new categories / section headings
  - String functions
  - Date functions
  - Number functions
  - Array functions
  - Map functions
  - Type Conversion functions
  - Miscellaneous functions (uncategorized multi-type functions)

- also consistently use `str`, `array`, `map`, `n`, `v` for the variable name when referring to a specific type
  - `array` was used consistently for older functions, but not some newer ones
  - `str` added to match `array`
  - `n` was sometimes used for numbers or ints, but not always
  - `v` is for multi-type "values"
  
## Validation

Read through in markdown previewer to double check

## Notes for Reviewers

Feel free to reject this if you'd rather keep it uncategorized! Categories are subjective, after all, so preferences can differ.
Also feel free to recommend other categorizations or re-categorizations!

### DIFF
--- a/docs/Language-Definition.md
+++ b/docs/Language-Definition.md
@@ -193,88 +193,6 @@ For example, expression `split(lower(user.Name), " ")` can be written as:
 user.Name | lower() | split(" ")
 ```
 
-## Type Conversion Functions
-
-### type(v)
-
-Returns the type of the given value `v`.
-Returns on of the following types: `nil`, `bool`, `int`, `uint`, `float`, `string`, `array`, `map`.
-For named types and structs, the type name is returned.
-
-```expr
-type(42) == "int"
-type("hello") == "string"
-type(now()) == "time.Time"
-```
-
-### int(v)
-
-Returns the integer value of a number or a string.
-
-```expr
-int("123") == 123
-```
-
-### float(v)
-
-Returns the float value of a number or a string.
-
-### string(v)
-
-Converts the given value `v` into a string representation.
-
-```expr
-string(123) == "123"
-```
-
-### toJSON(v)
-
-Converts the given value `v` to its JSON string representation.
-
-```expr
-toJSON({"name": "John", "age": 30})
-```
-
-### fromJSON(v)
-
-Parses the given JSON string `v` and returns the corresponding value.
-
-```expr
-fromJSON('{"name": "John", "age": 30}')
-```
-
-### toBase64(v)
-
-Encodes the string `v` into Base64 format.
-
-```expr
-toBase64("Hello World") == "SGVsbG8gV29ybGQ="
-```
-
-### fromBase64(v)
-
-Decodes the Base64 encoded string `v` back to its original form.
-
-```expr
-fromBase64("SGVsbG8gV29ybGQ=") == "Hello World"
-```
-
-### toPairs(map)
-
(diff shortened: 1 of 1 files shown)


docs/Language-Definition.md
```diff
@@ -205,460 +205,454 @@ date("2023-08-14") - date("2023-08-13") == duration("24h")
...
 ```
 
-### none(array, predicate)
-
-Returns **true** if _all elements does not_ satisfy the [predicate](#predicate).
-If the array is empty, returns **true**.
+#### Type Conversion Functions
 
-### map(array, predicate)
+#### type(v)
```
antonmedv, line 224: Lets keep function on third level `###`. This way they will appear in index on https://expr.medv.io/docs/Language-Definition.

docs/Language-Definition.md
```diff
@@ -205,460 +205,454 @@ date("2023-08-14") - date("2023-08-13") == duration("24h")
...
-one(participants, {.Winner})
+get([1, 2, 3], 1) == 2
+get({"name": "John", "age": 30}, "name") == "John"
 ```
 
-### none(array, predicate)
-
-Returns **true** if _all elements does not_ satisfy the [predicate](#predicate).
-If the array is empty, returns **true**.
+#### Type Conversion Functions
```
antonmedv, line 222: Let's keep  categories on the second level. `##`.

antonmedv (changes requested):

antonmedv:
I like the idea.

docs/Language-Definition.md
agilgur5: If Categories will be at the second level, should we remove the `## Built-in Functions` heading that these are all currently under?

docs/Language-Definition.md
agilgur5: yea I would like them to appear in the index too, wasn't sure what heading level it tracked until

docs/Language-Definition.md
agilgur5: Renamed it to "Miscellaneous Functions" as if everything else is an h2, this only has the misc functions in it. And I moved misc to the bottom of the functions list since they are miscellaneous

agilgur5:
made some adjustments per feedback

antonmedv:
Lets also rearrange order: string function, number funcs, array, date, type, misc.

agilgur5:
> Lets also rearrange order: string function, number funcs, array, date, type, misc.

where would you like to place map functions?

also I put date functions right after string because technically they are all strings as well

agilgur5:
I did string > date > number > array > map > type > misc. Let me know if that's alright or you want more changes

antonmedv (approved):

antonmedv:
Done! https://expr.medv.io/docs/Language-Definition

Thanks for the awesome contribution.

agilgur5:
new heading index looks great 😄 :
Screenshot 2023-10-04 at 5 37 51 PM

Issue #466, opened 2023-11-13
cannot fetch Value from *interface


a-peyrard:
First of all, thanks for you amazing library!

I'm facing some error when I'm using some dynamic unmarshalling of protobuf messages in the program context, the unmarshalled types are `interface{} | *interface{} | *<STRUCT CONCRETE TYPE>`.
When I'm trying to access a field of the struct, I'm getting the error: `cannot fetch Value from *interface{}`.

Here is a concrete example where I'm reproducing the error:
```go
func TestExpr_fetch_from_interface(t *testing.T) {
	type FooBar struct {
		Value string
	}
	foobar := &FooBar{"waldo"}
	var foobarAny any = foobar
	var foobarPtrAny any = &foobarAny

	res, err := expr.Eval("foo.Value", map[string]any{
		"foo": foobarPtrAny,
	})
	assert.NoError(t, err)
	assert.Equal(t, "waldo", res)
}
```
And the error is:
```
=== RUN   TestExpr_fetch_from_interface
    expr_test.go:1218: 
        	Error Trace:	/opt/dev/go/expr/expr_test.go:1218
        	Error:      	Received unexpected error:
        	            	cannot fetch Value from *interface {} (1:5)
        	            	 | foo.Value
        	            	 | ....^
        	Test:       	TestExpr_fetch_from_interface
    expr_test.go:1219: 
        	Error Trace:	/opt/dev/go/expr/expr_test.go:1219
        	Error:      	Not equal: 
        	            	expected: string("waldo")
        	            	actual  : <nil>(<nil>)
        	Test:       	TestExpr_fetch_from_interface
--- FAIL: TestExpr_fetch_from_interface (0.00s)
```

This was working fine with `v1.12.5` but this not working anymore with `v1.15.3`.

The error is triggered in the `Fetch` method, here: https://github.com/antonmedv/expr/blob/5916dc2b68d01606f6c23e1b54da5ee3cd3dc79a/vm/runtime/runtime.go#L76

I'm wondering a simple fix could be to use a deref function like that:
```go
func deref(kind reflect.Kind, value reflect.Value) (reflect.Kind, reflect.Value) {
	for kind == reflect.Ptr || kind == reflect.Interface {
		value = value.Elem()
		kind = value.Kind()
	}
	return kind, value
}
```
[...]

PR #467, opened 2023-11-14
Fix Fetch and OpDeref

a-peyrard:
First commit is fixing some issue with `runtime.Deref` which is called by `OpDeref`, pointer of interface were not de-referenced.
Second commit is creating a fix for #466. This is de-referencing everything in `Fetch` before fetching the attribute. 
As discussed in the issue, one other approach could be to ensure we have a `OpDeref` operation added during the compilation phase.

### DIFF
--- a/vm/runtime/runtime.go
+++ b/vm/runtime/runtime.go
@@ -145,25 +145,11 @@ func Deref(i any) any {
 
 	v := reflect.ValueOf(i)
 
-	if v.Kind() == reflect.Interface {
+	for v.Kind() == reflect.Ptr || v.Kind() == reflect.Interface {
 		if v.IsNil() {
-			return i
+			return nil
 		}
 		v = v.Elem()
-	}
-
-loop:
-	for v.Kind() == reflect.Ptr {
-		if v.IsNil() {
-			return i
-		}
-		indirect := reflect.Indirect(v)
-		switch indirect.Kind() {
-		case reflect.Struct, reflect.Map, reflect.Array, reflect.Slice:
-			break loop
-		default:
-			v = v.Elem()
-		}
 	}
 
 	if v.IsValid() {
--- a/vm/runtime/runtime.go
+++ b/vm/runtime/runtime.go
@@ -8,6 +8,14 @@ import (
 	"reflect"
 )
 
+func deref(kind reflect.Kind, value reflect.Value) (reflect.Kind, reflect.Value) {
+	for kind == reflect.Ptr || kind == reflect.Interface {
+		value = value.Elem()
+		kind = value.Kind()
+	}
+	return kind, value
+}
+
 func Fetch(from, i any) any {
 	v := reflect.ValueOf(from)
 	kind := v.Kind()
@@ -28,10 +36,8 @@ func Fetch(from, i any) any {
 	// Structs, maps, and slices can be access through a pointer or through
 	// a value, when they are accessed through a pointer we don't want to
 	// copy them to a value.
-	if kind == reflect.Ptr {
-		v = reflect.Indirect(v)
-		kind = v.Kind()
-	}
+	// De-reference everything if necessary (interface and pointers)
+	kind, v = deref(kind, v)
 
 	// TODO: We can create separate opcodes for each of the cases below to make
 	// the little bit faster.
--- a/expr_test.go
+++ b/expr_test.go
@@ -1204,6 +1204,21 @@ func TestExpr_fetch_from_func(t *testing.T) {
 	assert.Contains(t, err.Error(), "cannot fetch Value from func()")
 }
 
+func TestExpr_fetch_from_interface(t *testing.T) {
+	type FooBar struct {
+		Value string
+	}
+	foobar := &FooBar{"waldo"}
+	var foobarAny any = foobar
+	var foobarPtrAny any = &foobarAny
+
+	res, err := expr.Eval("foo.Value", map[string]any{
+		"foo": foobarPtrAny,
+	})
+	assert.NoError(t, err)
+	assert.Equal(t, "waldo", res)
+}
+
 func TestExpr_map_default_values(t *testing.T) {
 	env := map[string]any{
 		"foo": map[string]string{},
--- a/test/deref/deref_test.go
+++ b/test/deref/deref_test.go
@@ -137,3 +137,66 @@ func TestDeref_multiple_pointers(t *testing.T) {
 		require.Equal(t, 44, output)
 	})
 }
+
+func TestDeref_pointer_of_interface(t *testing.T) {
+	v := 42
+	a := &v
+	b := any(a)
+	c := any(&b)
+	t.Run("returned as is", func(t *testing.T) {
+		output, err := expr.Eval(`c`, map[string]any{
+			"c": c,
+		})
+		require.NoError(t, err)
+		require.Equal(t, c, output)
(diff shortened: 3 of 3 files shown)


a-peyrard:
I will have a look tomorrow to see if I can understand where I would need to modify the code to add the `OpDeref` operation. 
I would also like to check if the modifications are not impacting the overall performance as the two modifications are for functions used during runtime.

antonmedv:
Results looks [promising](https://github.com/antonmedv/expr/actions/runs/6860103470/job/18675207713?pr=467#step:8:11):
```
goos: linux
goarch: amd64
pkg: github.com/antonmedv/expr
cpu: AMD EPYC 7763 64-Core Processor                
                           │ /tmp/old.txt │            /tmp/new.txt            │
                           │    sec/op    │   sec/op     vs base               │
_expr-4                       119.2n ± 2%   113.1n ± 3%  -[5](https://github.com/antonmedv/expr/actions/runs/6860103470/job/18675207713?pr=467#step:8:6).12% (p=0.001 n=10)
_expr_reuseVm-4               72.57n ± 0%   71.27n ± 3%  -1.79% (p=0.022 n=10)
_len-4                        73.27n ± 1%   73.41n ± 1%       ~ (p=0.085 n=10)
_filter-4                     84.2[6](https://github.com/antonmedv/expr/actions/runs/6860103470/job/18675207713?pr=467#step:8:7)µ ± 1%   [7](https://github.com/antonmedv/expr/actions/runs/6860103470/job/18675207713?pr=467#step:8:8)9.42µ ± 1%  -5.74% (p=0.000 n=10)
_filterLen-4                  75.44µ ± 2%   74.12µ ± 2%  -1.75% (p=0.005 n=10)
_filterFirst-4                7[8](https://github.com/antonmedv/expr/actions/runs/6860103470/job/18675207713?pr=467#step:8:9)2.6n ± 1%   750.6n ± 1%  -4.10% (p=0.000 n=10)
_filterLast-4                 1.400µ ± 1%   1.341µ ± 0%  -4.25% (p=0.000 n=10)
_filterMap-4                  [9](https://github.com/antonmedv/expr/actions/runs/6860103470/job/18675207713?pr=467#step:8:10).045µ ± 0%   8.785µ ± 1%  -2.87% (p=0.000 n=[10](https://github.com/antonmedv/expr/actions/runs/6860103470/job/18675207713?pr=467#step:8:11))
_arrayIndex-4                 [11](https://github.com/antonmedv/expr/actions/runs/6860103470/job/18675207713?pr=467#step:8:12)3.5n ± 0%   117.1n ± 0%  +3.22% (p=0.000 n=10)
_envStruct-4                  85.53n ± 0%   85.89n ± 2%  +0.42% (p=0.014 n=10)
_envMap-4                     96.50n ± 2%   96.47n ± 2%       ~ (p=0.566 n=10)
_callFunc-4                   617.1n ± 0%   620.6n ± 1%  +0.58% (p=0.003 n=10)
_callMethod-4                 638.2n ± 0%   653.5n ± 1%  +2.40% (p=0.000 n=10)
_callField-4                  115.2n ± 0%   1[13](https://github.com/antonmedv/expr/actions/runs/6860103470/job/18675207713?pr=467#step:8:14).8n ± 1%  -1.22% (p=0.001 n=10)
_callFast-4                   119.1n ± 0%   115.1n ± 0%  -3.44% (p=0.000 n=10)
_callConstExpr-4              93.50n ± 1%   93.50n ± 1%       ~ (p=0.699 n=10)
_largeStructAccess-4          227.8n ± 1%   231.5n ± 1%  +1.67% (p=0.001 n=10)
_largeNestedStructAccess-4    239.5n ± 1%   240.6n ± 1%  +0.46% (p=0.017 n=10)
_largeNestedArrayAccess-4     1.355m ± 4%   1.348m ± 2%       ~ (p=0.481 n=10)
_sort-4                       7.721µ ± 2%   7.820µ ± 1%  +1.28% (p=0.043 n=10)
_sortBy-4                     15.00µ ± 1%   [14](https://github.com/antonmedv/expr/actions/runs/6860103470/job/18675207713?pr=467#step:8:15).94µ ± 0%  -0.43% (p=0.004 n=10)
_groupBy-4                    17.97µ ± 0%   17.61µ ± 0%  -2.01% (p=0.000 n=10)
_reduce-4                     8.539µ ± 0%   8.104µ ± 1%  -5.08% (p=0.000 n=10)
geomean                       1.[15](https://github.com/antonmedv/expr/actions/runs/6860103470/job/18675207713?pr=467#step:8:16)0µ        1.135µ       -1.[26](https://github.com/antonmedv/expr/actions/runs/6860103470/job/18675207713?pr=467#step:8:27)%
```

vm/runtime/runtime.go
```diff
@@ -28,10 +36,8 @@ func Fetch(from, i any) any {
 	// Structs, maps, and slices can be access through a pointer or through
 	// a value, when they are accessed through a pointer we don't want to
 	// copy them to a value.
-	if kind == reflect.Ptr {
-		v = reflect.Indirect(v)
```
antonmedv, line 32: Looks like we going to loose `reflect.Indirect` here.

antonmedv: Here is ooriginal change: https://github.com/antonmedv/expr/commit/6f0c8555bad3d5ab09f35c1ef3c3047382514e3c

a-peyrard: We are still doing the "indirect" in the `deref` function as we are calling `Elem()` of the value is a pointer.
Which should be doing the same as the `Indirect` function, which is also doing `Elem()` behind the scene:
```go
func Indirect(v Value) Value {
	if v.Kind() != Pointer {
		return v
	}
	return v.Elem()
}
```

a-peyrard: I just ran the benchrmak, it seems we are in the same order of magnitude:
before: 
```
Benchmark_largeStructAccess-12           1000000               134.4 ns/op
```
after:
```
Benchmark_largeStructAccess-12           1000000               132.2 ns/op
```

a-peyrard:
I was trying to figure out where we would need to add an `OpDeref` operation. I was thinking to add one there:
https://github.com/antonmedv/expr/blob/709c5dd55aa7cfc1b47fc67ee32ac15af6db967a/compiler/compiler.go#L590
like that:
```go
if op == OpFetch {
		c.derefInNeeded(base)
		c.compile(node.Property)
		c.emit(OpFetch)
	}
```
This is fixing the problem when we try to access a field, but introducing a bug when fetch is used to get a method. This test for example is failing: `TestDeref_method_on_int_pointer` as the element has been dereferenced and then doesn't contain any method named `Bar`.

Any idea where I should add a `OpDeref`, as `Fetch` is used to fetch fields and methods, it might be a bit tricky w/o splitting the memberNode into two separate structures, one to retrieve methods, and one to retrieve fields.

vm/runtime/runtime.go
antonmedv: Cool! let me check it a bit more as runtime changes should be done very carefully.

antonmedv:
> The other solution would be to not add any OpDeref and just do the deref if needed during the Fetch (as implemented in this PR)

Let's use this for now.

antonmedv:
> The other solution would be to not add any OpDeref and just do the deref if needed during the Fetch (as implemented in this PR)

Lets use this for now.

antonmedv (approved):

PR #470, opened 2023-11-17
Implement `In` operator for string and string

filipecaixeta:
Implemented `In` operator for string and string
This will allow expressions like
```
'a' in 'abc' 
```


antonmedv:
We already have this one:

    'abc' contains 'a'

PR #485, opened 2023-12-01
Incorporate support for raw string literals

ckganesan:
The current parser lacks support for raw string literals

### DIFF
--- a/parser/lexer/lexer.go
+++ b/parser/lexer/lexer.go
@@ -219,3 +219,17 @@ func (l *lexer) scanString(quote rune) (n int) {
 	}
 	return
 }
+
+func (l *lexer) scanRawString(quote rune) (n int) {
+	ch := l.next() // read character after back tick
+	for ch != quote {
+		if ch == eof {
+			l.error("literal not terminated")
+			return
+		}
+		ch = l.next()
+		n++
+	}
+	l.emitValue(String, l.input[l.start+1:l.end-1])
+	return
+}
--- a/parser/lexer/lexer_test.go
+++ b/parser/lexer/lexer_test.go
@@ -48,6 +48,19 @@ func TestLex(t *testing.T) {
 			},
 		},
 		{
+			"`backtick` `hello\u263Aworld` `hello\n\tworld` `hello\"world'`  `\xC3\xBF\u263A\U000003A8` `❤️`",
+			[]Token{
+				{Kind: String, Value: `backtick`},
+				{Kind: String, Value: `hello☺world`},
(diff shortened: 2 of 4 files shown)


antonmedv:
Nice! But I'm thinking about adding byte strings later as well.
 
    b'bytes string'

So maybe it is better to add raw strings was a prefix?

    r'raw string\n'
    r"any quotes"
    r`and those too`

But allow back quotas to be multi line?

    r`nice
    multi
    line`

ckganesan:
I suggest that the `r` prefix be reserved specifically for raw strings enclosed in single (`' '`) or double quotes (`" "`), as this approach clearly distinguishes between standard and raw strings. For raw strings without any prefix, we can utilize backticks (``)

ckganesan:
If you aim to implement raw strings with the `r` prefix for both single (`' '`) and double quotes (`" "`). I will take on this task and submit a pull request within the same PR.

antonmedv:
> I recommend that we reserve the r prefix exclusively for raw strings enclosed in single (' ') or double quotes (" "), aligning with your previous comment for a clear distinction between standard and raw strings. Additionally, for raw strings that do not require a prefix, using backticks (``) would be an effective approach.

What is reason?

ckganesan:
> > I recommend that we reserve the r prefix exclusively for raw strings enclosed in single (' ') or double quotes (" "), aligning with your previous comment for a clear distinction between standard and raw strings. Additionally, for raw strings that do not require a prefix, using backticks (``) would be an effective approach.
> 
> What is reason?

Backticks are inherently interpreted as raw strings, which means there's no need to use escape sequences. However, when using quotes to implement raw strings, escape sequences become essential.

ckganesan:
That's why I believe there is no need for the `r` prefix when using backtick symbols alone.

antonmedv:
I see. This makes good point. 

I was thinking about how raw string implemented in python.

ckganesan:
While Python uses the r prefix for raw strings, Go achieves this using backtick symbols.

```go
// Go Using backticks for raw string
rawString := `This is a raw string in Go.
It can span multiple lines and include "quotes" without escaping them.`
```

```python
# python Using the 'r' prefix for raw string
raw_string = r"This is a raw string in Python.\nIt can include escape characters like \n, \t, etc., without any special effect."
```

We can achieve raw strings in two ways: using the r prefix or with backticks

antonmedv:
Ok. Let me review pr more carefully.

antonmedv (approved):

PR #487, opened 2023-12-01
Patcher and interfaces for custom types to be used in expressions as standard go types

rrb3942:
See https://github.com/antonmedv/expr/pull/460 for previous discussion.

I tried to cover all the types supported by expr, as well generic `map[string]any` and `[]any`.

To take advantage of the interfaces you use `value.Patcher` as an option when compiling an expression.


patchers/value/value.go
```diff
@@ -0,0 +1,260 @@
...
+
+type patcher struct{}
+
+func (patcher) Visit(node *ast.Node) {
+	id, ok := (*node).(*ast.IdentifierNode)
+	if !ok {
+		return
+	}
+
+	callnode := &ast.CallNode{
```
antonmedv, line 176: Let's create CallNode only if we need to (e.g. inside if statment).

patchers/value/value.go
```diff
@@ -0,0 +1,260 @@
...
+		Callee:    &ast.IdentifierNode{Value: "getExprValue"},
+		Arguments: []ast.Node{id},
+	}
+
+	nodeType := id.Type()
+
+	for _, t := range supportedInterfaces {
+		if nodeType.Implements(t) {
```
antonmedv, line 184: What going to be if the type `t` is nil?

patchers/value/value.go
```diff
@@ -0,0 +1,260 @@
...
+		}
+	}
+}
+
+func (patcher) ApplyOptions(c *conf.Config) {
+	getExprValueFunc(c)
+}
+
+func getExprValue(params ...any) (any, error) {
+
```
antonmedv, line 195: code style: extra space.

patchers/value/value.go
```diff
@@ -0,0 +1,260 @@
...
+	"time"
+
+	"github.com/antonmedv/expr"
+	"github.com/antonmedv/expr/ast"
+	"github.com/antonmedv/expr/conf"
+)
+
+// Patcher is an expr.Option that both patches the program and adds the `getExprValue` function.
+// Use it directly as an Option to expr.Compile()
+var Patcher = func() expr.Option {
```
antonmedv, line 57: Let's use patcher pkg and rename Patcher to, for example, `patcher.ValueGetter`.

patchers/value/value.go
```diff
@@ -0,0 +1,260 @@
...
+	vPatcher := patcher{}
+	return func(c *conf.Config) {
```
antonmedv, line 59: We can reuse `expr.Patch` helper here.

patchers/value/value.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+func (patcher) ApplyOptions(c *conf.Config) {
+	getExprValueFunc(c)
+}
+
+func getExprValue(params ...any) (any, error) {
+
+	switch v := params[0].(type) {
+	case ExprValuer:
```
antonmedv, line 197: What about creating a separate function for each type?

patchers/value/value.go
```diff
@@ -0,0 +1,260 @@
...
+	case ArrayValuer:
+		return v.ArrayValue(), nil
+	case MapValuer:
+		return v.MapValue(), nil
+	}
+
+	return params[0], nil
+}
+
+var getExprValueFunc = expr.Function("getExprValue", getExprValue,
```
antonmedv, line 240: Also let's rename `getExprValue` to something what will definitely not collide with user defined functions.

Like, `$patcher_value_getter`.

patchers/value/value.go
```diff
@@ -0,0 +1,260 @@
+// Package value provides a Patcher that uses interfaces to allow custom types that can be represented as standard go values to be used more easily in expressions.
+//
+// # Example Usage
```
antonmedv, line 3: Let's add documentation about motivation and there this patcher can be used.

antonmedv (changes requested):

patchers/value/bench_test.go
```diff
@@ -0,0 +1,71 @@
+package value
+
+import (
+	"github.com/antonmedv/expr"
+	"github.com/antonmedv/expr/vm"
+	"github.com/stretchr/testify/require"
+	"testing"
```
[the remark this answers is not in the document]
rrb3942: fe53377e288726e9fc51e58caacb6d5e76ed3a38

patchers/value/value.go
rrb3942: b1b1b84589a68a63343600fdd2c814cb3fcd9826

patchers/value/value.go
rrb3942: fe53377e288726e9fc51e58caacb6d5e76ed3a38

patchers/value/value.go
rrb3942: c9797215aab8a4f000e1eeb1c00edf660c2b8592

patchers/value/value.go
```diff
@@ -0,0 +1,260 @@
...
+// It allows for returning a `nil` value but does not provide any type checking at expression compile.
+//
+// A custom type may implement both ExprValuer and a type specific interface to enable both
+// compile time checking and the ability to return a `nil` value.
+type ExprValuer interface {
+	ExprValue() any
+}
+
+type IntValuer interface {
+	IntValue() int
```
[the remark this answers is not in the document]
rrb3942: I'm not happy with the `Value` suffix either. The big issue is running into collisions.

Example of unwanted collisions:
`big.Float` - has `Int`, `Int64`, `Float32`, `Float64`, `Uint64`, and `String` as methods already. Signatures are mostly different, but I think it is reasonable someone might want to embed it and add and interface for `expr` use.  Complexity goes way up if you are all of a sudden dealing with naming collisions.

`fmt.Stringer` - `String` method. To start `time.Time` implements this, so special casing would be required or we would turn all times into strings. I also don't think it's unreasonable for a custom type to implement `String` for printing, but want to use a different interface with `expr`. Multiple matching interfaces leads to the question of precedence, which I would like to mostly avoid, except for the `Any` interface with a specific type interface.

`Value` suffix actually causes issues in my own project, since I use https://pkg.go.dev/github.com/jackc/pgx/v5@v5.5.0/pgtype 

What about an `As` prefix? `AsAny()`, `AsInt()`? Or something else entirely?

patchers/value/value.go
rrb3942: Would this be for performance reasons? I didn't want to pollute the `expr` function space with a bunch of functions.

I want to preserve the ability to implement both `Any` and a specific type interface. This lets you get compile type checks for the type, but still return `nil` at runtime. So, even if specific functions for each type were implemented, a type check would still need to be done. In some quick testing I had done earlier it seemed like as soon as you were paying the cost of type checking it didn't really matter the number of branches in the switch, or if you broke it into an `if` statement.

I can revisit if you would like.

patchers/value/value.go
rrb3942: `t` should never be nil.

`supportedInterfaces` is a slice containing all supported interface types and should never contain a `nil`.

patchers/value/bench_test.go
```diff
@@ -0,0 +1,71 @@
+package value
```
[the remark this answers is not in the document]
rrb3942: Of course. I wasn't sure where to put it.

patchers/value/value.go
rrb3942: How should I define the actual expr.Function doing this? Should I move it somewhere else?

`expr` imports `patcher` so it creates an import cycle to create the function in `patcher`.

patchers/value/value.go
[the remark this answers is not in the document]
antonmedv: Good point. 

Yes `As` looks like a good short alternative.

patchers/value/value.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+}()
+
+// A ExprValuer provides a generic function for a custom type to return standard go values.
+// It allows for returning a `nil` value but does not provide any type checking at expression compile.
+//
+// A custom type may implement both ExprValuer and a type specific interface to enable both
+// compile time checking and the ability to return a `nil` value.
+type ExprValuer interface {
+	ExprValue() any
```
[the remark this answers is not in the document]
rrb3942: a7e4934bf2f579623ebe06f49c2139a716e3697e

patchers/value/value.go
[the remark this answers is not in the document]
rrb3942: a7e4934bf2f579623ebe06f49c2139a716e3697e

patchers/value/value.go
rrb3942: @antonmedv

Any advice on where I should define the expr.Function without causing an import cycle?

I don't think it should be defined as a normal builtin function?

antonmedv: Sorry, I didn't understand the question. 

I mean this: lets move code to patchers dir and rename this function to ValueGetter.

Issue #207, opened 2021-11-09
Add support for bitwise operators


ShengXiangXiong:
```
val := `1 & 1`
complie, err := expr.Compile(val)
r, err := expr.Run(complie,env)
if err!=nil {
fmt.Printf("%+v", err)
}
if v, ok := r.(int); ok {
fmt.Printf("%+v", v)
}
```

![image](https://user-images.githubusercontent.com/20365712/140894195-1b6d9a69-b02e-426b-abef-ee08002ee9de.png)

antonmedv:
Nope, no bit operators.

ShengXiangXiong:
Can this feature be added in the future

ShengXiangXiong:
> Nope, no bit operators.

Can this feature be added in the future

antonmedv:
Maybe.

antonmedv:
I'm thinking to add bit ops with such functions:

- shl(int, int) – signed shift left
- shr(int, int) – signed shift right
- ushr(int, int) – unsigned shift right
- band(int, int) – bitwise AND
- bor(int, int) – bitwise OR
- xor(int) – bitwise XOR
- inv(int) – bitwise inversion

antonmedv:
Now Expr supports bit operations. See https://expr-lang.org/docs/language-definition#bitwise-functions

PR #496, opened 2023-12-11
feat: Implement Bitwise and Shift Operator Support

ckganesan:
This PR introduces support for bitwise and shift operators. With this enhancement, users can now perform advanced bitwise operations, including AND, OR, XOR, and NOT, as well as left and right shift operations.

```
&    bitwise AND            integers
// To manage bitwise OR operations and various callbacks (predicates, built-ins, and custom functions)
|    bitwise OR             integers
// Duplicate the Caret operator to prevent confusion with the Exponent operator.
^^    bitwise XOR           integers
&^   bit clear (AND NOT)    integers

<<   left shift             integer << integer >= 0
>>   right shift            integer >> integer >= 0
```

Ref: #207


antonmedv:
Let’s not override pipe operator. This will create confusion. Also ^^ is really odd. 

What about using full names as operators for bitwise ops?

I already know a good one:

- xor


    A xor B

We just need to find nice ones for other. I know some languages do that.

antonmedv:
And let’s implement them as functions. So we don’t need to worry about precedence.

antonmedv:
Or as alternative solution add a special "macro": `bit` which will be converting those expressions:

```js
bit(A | B ^ ~C) 
```

ckganesan:
> And let’s implement them as functions. So we don’t need to worry about precedence.

Following the comments, I will also be adding the following built-in functions, along with their respective precedences.

```
// Bitwise ops
&, bitand     -    bitwise AND            integers
bor, bitor    -    bitwise OR             integers
xor, bitxor   -    bitwise XOR            integers
&^, bitnand   -    bit clear (AND NOT)    integers
~, bitnot     -    bit NOT                integer

// Bit Shift ops
<<, bitshl    -    left shift             integer << integer >= 0
>>, bitshr    -    right shift            integer >> integer >= 0
>>>, bitushr  -    unsigned right shift 
```

antonmedv:
But let's drop operators

PR #499, opened 2023-12-13
Add duration literal support

fmarcia:
This pull request adds support for duration literal:
```expr
-1h == -1 * 1h
+1h == 1h
2 * 1h == 2h
1h + 1m == 1h1m
1h - 1m == 59m
1h / 10m == 6
1h / 2 == 30m
```
round, max, min and abs builtin functions have been updated to support duration as well.


antonmedv:
I'm not sure yet if this should be added directly to the language itself. We already have duration() func.

Why do you think this should be implemented directly in the language?

fmarcia:
Mostly for the sake of readability. I'm using durations frequently, and I find literal format more readable and quicker to type as well.

antonmedv:
For now, I desiced it is better to stick with classic `duration()` function. With it it is possible to do neat tricks:

```js
duration("2h").Minutes()
```

antonmedv:
For now, I desiced it is better to stick with classic `ducration()` function. With it it is possible to do neat tricks:

PR #500, opened 2023-12-18
Fix: checker panic on embedded pointer to struct field

sgreben:
Here's a fix for a case I ran into in practice - the checker panics on a valid input:

expression: 
```go
(Embed).EmbedPointerEmbedInt > 0
```
environment: 
```go
type Env struct {
	Embed
}
type Embed struct {
	*EmbedPointerEmbed
}

type EmbedPointerEmbed struct {
	EmbedPointerEmbedInt int
}
```

crash:
```
panic: reflect: NumField of non-struct type *mock.EmbedPointerEmbed [recovered]
        panic: reflect: NumField of non-struct type *mock.EmbedPointerEmbed
```

The fix in this PR is to dereference embedded fields until they stop being pointers (see diff in checker/types.go).

### DIFF
--- a/checker/checker_test.go
+++ b/checker/checker_test.go
@@ -131,6 +131,7 @@ func TestCheck(t *testing.T) {
 		{"(Any.Bool ?? Bool) > 0"},
 		{"Bool ?? Bool"},
 		{"let foo = 1; foo == 1"},
+		{"(Embed).EmbedPointerEmbedInt > 0"},
 	}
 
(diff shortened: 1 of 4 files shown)


antonmedv:
Let’s also add a rest what evaluation works and we can call a method on EmbeddedPointerStruct.

sgreben:
^added (passing) tests for eval & method calls

PR #505, opened 2023-12-29
Add mem grow to builtin

antonmedv:


antonmedv:
Screenshot 2023-12-30 at 10 30 46
Screenshot 2023-12-30 at 10 30 53
Original implementation was slow.

antonmedv:
New is good.

PR #506, opened 2024-01-02
Introduced a built-in `deref` function specifically designed to recursively dereference any provided value.

ckganesan:
This PR introduces the `deref` function, designed to efficiently dereference given values across multiple levels of pointers, enhancing value accessibility in our expr.

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -1028,4 +1028,9 @@ var Builtins = []*Function{
 		},
 		Types: types(new(func(int) int)),
 	},
+	{
+		Name:  "deref",
+		Fast:  Deref,
+		Types: types(new(func(any) any)),
+	},
 }
--- a/builtin/builtin_test.go
+++ b/builtin/builtin_test.go
@@ -19,11 +19,16 @@ import (
 )
 
 func TestBuiltin(t *testing.T) {
+	a := "derefTest"
+	ptrA := &a
+	ptrPtrA := &ptrA
 	env := map[string]any{
 		"ArrayOfString": []string{"foo", "bar", "baz"},
 		"ArrayOfInt":    []int{1, 2, 3},
 		"ArrayOfAny":    []any{1, "2", true},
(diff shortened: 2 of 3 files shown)


antonmedv:
Can you show a use case? Right now we have bytecode **OpDeref** which is automatically added and does exactly the same thing.

ckganesan:
> Can you show a use case? Right now we have bytecode **OpDeref** which is automatically added and does exactly the same thing.

The OpDeref bytecode is exclusively supported for operators, and it does not extend to other elements like functions or chain nodes. Please find the below example

```go
	a := "Hello"
	ptrA := &a
	ptrPtrA := &ptrA

	code := `lower(ptrA)`
		"ptrA":    ptrA,
		"ptrPtrA": ptrPtrA,
	}

	if err != nil {
		panic(err)
	}

	if err != nil {
		panic(err)
	}
	fmt.Print(output)
```
```
panic: cannot use *string as argument (type string) to call lower  (1:7)
 | lower(ptrA)
 | ......^
```

After changing the above code like below

```go
	code := `lower(deref(ptrA))` // output 'hello'
        code := `deref(ptrPtrA) | lower()` // output 'hello'
        code := `len(deref(ptrPtrA))` // output 5
```

This function seamlessly handles single or multiple pointers by dereferencing the value regardless of its pointer depth.

ckganesan:
The `deref` function provides flexibility for users familiar with their expression environment objects, allowing them to selectively invoke dereferencing only when necessary. This approach optimizes performance by avoiding unnecessary dereferencing across all operators, regardless of whether a pointer is present.

antonmedv:
What about adding OpDeref opcode for arguments if needed? This way your example with string will work. 

I was thinking about adding deref() myself, but decided what it will be easier for inexperienced users to do all the dereference via opcodes.

ckganesan:
> What about adding OpDeref opcode for arguments if needed? This way your example with string will work.
> 
> I was thinking about adding deref() myself, but decided what it will be easier for inexperienced users to do all the dereference via opcodes.

Got it, i will update as per your comments

PR #523, opened 2024-01-13
Fix race cond in variabels

### LINKED ISSUE #519: OpStore modifies the Program (but shouldn't)
Run writes to `program.variables` for `OpStore`:

https://github.com/expr-lang/expr/blob/1b3c10c0ef680ae4e6b6b3558308b233958b07f2/vm/vm.go#L114-L115

...which is a data race if the `Program` is shared between goroutines. Should `variables` be a `VM` field instead?

antonmedv:
Fixes #519

### DIFF
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -78,7 +78,7 @@ func (vm *VM) Run(program *Program, env any) (_ any, err error) {
 	if vm.Scopes != nil {
 		vm.Scopes = vm.Scopes[0:0]
 	}
-	if vm.Variables == nil {
+	if len(vm.Variables) < program.variables {
 		vm.Variables = make([]any, program.variables)
 	}
 
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -34,6 +34,7 @@ func Debug() *VM {
 type VM struct {
 	Stack        []any
 	Scopes       []*Scope
+	Variables    []any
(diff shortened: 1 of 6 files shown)


vm/vm.go
```diff
@@ -74,10 +75,12 @@ func (vm *VM) Run(program *Program, env any) (_ any, err error) {
 	} else {
 		vm.Stack = vm.Stack[0:0]
 	}
-
 	if vm.Scopes != nil {
 		vm.Scopes = vm.Scopes[0:0]
 	}
+	if vm.Variables == nil {
+		vm.Variables = make([]any, program.variables)
```
sgreben, line 82: nevermind if you're still working on it, but since I'm on the topic - should the VM be reusable for different programs? 

because then i think we need to resize the variables slice here depending on the number of variables in the current program, something like

```go
if cap(vm.Variables) < len(program.variables) { 
    // we need to allocate more in this case
    vm.Variables = make([]any, len(program.variables)) 
} else {
    // we can resize the already-allocated slice
    vm.Variables = vm.Variables[:len(program.variables)] 
}
```

antonmedv: Correct suggestion. Usually it make sense to only run a vm for a specific program (as we this way stack will be reallocated and potently will not grow)

PR #541, opened 2024-01-30
Split compile into create and compile steps.

AndrewDK:
This PR splits expr.Compile into a CreateConfig step and a CompileConfig step so that a Config can be reused across expressions.

### DIFF
--- a/expr.go
+++ b/expr.go
@@ -177,8 +177,8 @@ func WithContext(name string) Option {
 	})
 }
 
-// Compile parses and compiles given input expression to bytecode program.
-func Compile(input string, ops ...Option) (*vm.Program, error) {
+// CreateConfig creates a Config and applies the Options.
+func CreateConfig(ops ...Option) (*conf.Config, error) {
 	config := conf.CreateNew()
 	for _, op := range ops {
 		op(config)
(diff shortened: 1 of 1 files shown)


PranavPeshwe:
Do we need tests for the new functions/functionality?

AndrewDK:
> Do we need tests for the new functions/functionality?

Existing tests for expr.Config should demonstrate that there are no single "threaded" regressions, and that expr.CreateConfig and expr.CompileConfig work as expected.  However it doesn't guarantee that the Config can safely be used concurrently by multiple programs, which can now happen. I can add a test for that, but it requires an implementation contract/guarantee that the Config can never be updated by Compile. That seems to be the case now, but Is it reasonable to expect that going forward? The Config object is 297MB and creating it for every expression seems wasteful, so it seems like a good goal to reuse it, certainly since it is effectively static at this point.

AndrewDK:
I'm going to withdraw this. The performance improvements evaporate with large enough data, and don't justify the risk. It still seems like a good idea, but I'm not familiar enough with the internals to make sure it's safe.

antonmedv:
Not sure what this is needed. You can already reuse configuration:

```
options := []expr.Option{
    env.Env(),
    env.Function(),
    ....
}
`

antonmedv:
> every expr.Compile call still does a config := conf.CreateNew() and then applies those Options with the new Config, which is what I want to avoid.

Why? This is very fast. It is definitely not a place to look for optimizations. Compile can be slow. Focus on expr.Run to be fast.

Issue #552, opened 2024-02-08
Sort issue


PxyUp:
Hello @antonmedv !

First of all thanks for this one of the best tool which i know!

Here my problem(i am not sure is it problem or suggestion).

I trying sort array here [https://expr-lang.org/playground](https://expr-lang.org/playground):

```
sort(Posts) // or sort(Posts, 'desc')
```

From my point of view like user i am expect results should array with same order/reverse order here the error (as i understand no way to do that right now):

```
cannot deref <invalid reflect.Value> (3:1)
 | sort(Posts)
 | ^
 ```
 
 My suggestion is:  create new function which will be do that: `sortByIndex(array, order = 'asc|desc')` or create `reverse`. I am prefer for `reverse`(but may be need support string as well)

 Examples:
 
 ```
sortByIndex([3,2,1)) == [3,2,1]
sortByIndex([3,2,1), 'desc') == [1,2,3]
reverse([3,2,1]) == [1,2,3]
reverse(reverse([3,2,1])) == [3,2,1]
 ```
 
 I can create pull request with this function

**Comment by antonmedv** [comment]
Yes, [sort()](https://expr-lang.org/docs/language-definition#sort) is just for array on numbers, and sortBy() is for array structs/maps. 

```
sortBy(Posts, 'Title') | map(.ID)
```

And sure, lets add reverse().

PxyUp:
Adding MR: https://github.com/expr-lang/expr/pull/553

antonmedv:
Added #558

PR #553, opened 2024-02-08
code(builtin) - add reverse function

PxyUp:
Here pull request on the https://github.com/expr-lang/expr/issues/552

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -869,6 +869,40 @@ var Builtins = []*Function{
 		},
 	},
 	{
+		Name: "reverse",
+		Func: func(args ...any) (any, error) {
+			if len(args) != 1 {
+				return nil, fmt.Errorf("invalid number of arguments (expected 1, got %d)", len(args))
+			}
+
+			v := reflect.ValueOf(args[0])
+			if v.Kind() != reflect.Slice && v.Kind() != reflect.Array {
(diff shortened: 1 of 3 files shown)


builtin/builtin.go
```diff
@@ -868,6 +868,40 @@ var Builtins = []*Function{
...
+
+		},
+		Validate: func(args []reflect.Type) (reflect.Type, error) {
+			if len(args) != 1 {
+				return anyType, fmt.Errorf("invalid number of arguments (expected 1, got %d)", len(args))
+			}
+			switch kind(args[0]) {
+			case reflect.Interface, reflect.Slice, reflect.Array:
+			default:
+				return anyType, fmt.Errorf("cannot reverse %s", args[0])
```
antonmedv, line 900: Let's move this return inside `case`.

antonmedv (changes requested):
Please, add documentation to language-definition.md.

builtin/builtin.go
PxyUp: What you mean? you still will need return from the function

antonmedv: Nope.

```go
// You can edit this code!
// Click here and start typing.
package main

import "fmt"

func main() {
	fmt.Println(foo(42))
}

func foo(i int) bool {
	switch i {
	case 42:
		return true
	default:
		return false
	}
}
```
https://go.dev/play/p/YtfTFf9GjZ9

PR #555, opened 2024-02-09
Optimize boolean operations between all, any, one, none functions

bizywizy:


antonmedv:
Apparently we have a bug here:
```
let arr = [1, 2];
one(arr, # == 1) && one(arr, # == 2)
```

antonmedv:
Reported #624

PR #576, opened 2024-02-23
Enhanced min and max built-in functions to Support Arrays of Integers and Floats

ckganesan:
Enhanced support for array inputs in the min and max built-in functions, now accommodating arrays of integers and floats. This update extends the functionality to handle diverse data types seamlessly, improving the flexibility and usability of the functions. #574


antonmedv:
Cool! What about adding support for this as well?

```
max([1, 2, 3], 10) // output 10
```

antonmedv:
But I think it will complicate things. We can add `flatten()` builtin.

ckganesan:
> Cool! What about adding support for this as well?
> 
> ```
> max([1, 2, 3], 10) // output 10
> ```

Certainly, I can include that as well if you'd like.

antonmedv (approved):

antonmedv:
BTW, lets add this to https://expr-lang.org/docs/language-definition#mean and https://expr-lang.org/docs/language-definition#mean as well

PranavPeshwe:
Would be good to include some negative numbers in the tests.

Issue #572, opened 2024-02-20
Allow to use operator tokens as functions `contains()`


Q16G:
Could it be possible to add and disable an Operator and rewrite it entirely with function

Q16G:
What I mean by that is to rewrite the operand with a function after the change is disabled, and to re-specify the logic, for example, contains, it used to be "a" contains "b" can it now be "contains" ("a","b")

antonmedv:
I think allowing to use some operators as function is a bad practice. So I'm going to say not to this.

Issue #568, opened 2024-02-18
Add a method to disable Operator to change the usage logic of Operator


Q16G:
Whether to set that you can add a disabled Operator or disable an Operator? Because I want to rewrite the use of contains as follows: contains("abc","a"), but I find that after the rewrite, it is still used in the way of Operator and cannot be completely overwritten.

### code 

```
package operation

import (
	"fmt"
	"github.com/expr-lang/expr"
	"testing"
)

type Env struct {
	Contains func(_ string, _ string) bool
}

func TestOperations_Execute(t *testing.T) {
	var a = []expr.Option{
		expr.Env(Env{}),
		expr.DisableAllBuiltins(),
		expr.Operator("contains", "Contains"),
		expr.Function("Contains", func(params ...any) (any, error) {
			fmt.Println(params)
			return false, nil
		}),
	}
	compile, err := expr.Compile(`contains("a","a")`, a...)

	if err != nil {
		fmt.Println(err)
		return
	}
	run, err := expr.Run(compile, a)
	if err != nil {
		fmt.Println(err)
		return
	}
	fmt.Println(run)
}

```
#### output：
```
unexpected token Operator("contains") (1:1)
 | contains("a","a")
 | ^
```

antonmedv:
Here is a working example.  The problem is with expr expression. `contains` is an operator and can't be used as a function name (I can see if it is possible to change parser to support this use case as well).

Here is a correct example:

```js
"a" contains "b"
```

Also, a bug was fixed: now operator overloading can be used with expr.Function. 

```go
package main

import (
	"fmt"

	"github.com/expr-lang/expr"
)

func main() {
	options := []expr.Option{
		expr.Operator("contains", "Contains"),
		expr.Function(
			"Contains",
			func(params ...any) (any, error) {
				fmt.Println(params)
				return false, nil
			},
			new(func(string, string) bool),
		),
	}

	compile, err := expr.Compile(`"a" contains "b"`, options...)
	if err != nil {
		fmt.Println(err)
		return
	}

	run, err := expr.Run(compile, nil)
	if err != nil {
		fmt.Println(err)
		return
	}

	fmt.Println(run)
}
```

Q16G:
So can I disable the contains operator and then re-declare it as a function to use

PR #577, opened 2024-02-23
Allow to use operator tokens as functions

ckganesan:
This PR enables the usage of operator tokens as functions. #572


antonmedv:
Lets just allow to use "endsWith"
"startsWith"
"contains"
"matches" words as method names?

ckganesan:
> Lets just allow to use "endsWith" "startsWith" "contains" "matches" words as method names?

I have made the necessary updates based on your feedback.

Q16G:
This function is very important for my code, could you please add this function to the latest version with a tag, so that I can add it to the code

antonmedv:
Let's try to implement this feature without adding additional options. I will try to do it.

ckganesan:
> Let's try to implement this feature without adding additional options. I will try to do it.

I have option functionality from configuration as per your feedback

parser/parser_test.go
```diff
@@ -342,12 +349,24 @@ world`},
...
+				Right: &StringNode{Value: "foo"}},
+		},
 		{
 			`foo endsWith "foo"`,
 			&BinaryNode{Operator: "endsWith",
 				Left:  &IdentifierNode{Value: "foo"},
 				Right: &StringNode{Value: "foo"}},
 		},
+		{
+			`endsWith(foo,  "foo")`,
```
antonmedv, line 365: Let's also add test for postfix/prefix and other combinations as:
```
endsWith() endsWith endsWith()
endsWith().Field
endsWith()[0]
-+endsWith()
endsWith()+endsWith()
```
To see how parser whould behaive in complex scenarios?

parser/parser_test.go
```diff
@@ -342,12 +349,24 @@ world`},
...
+			&BinaryNode{Operator: "endsWith",
+				Left:  &IdentifierNode{Value: "foo"},
+				Right: &StringNode{Value: "foo"}},
+		},
```
antonmedv, line 369: And lets also add a test withing expr_test to see if it is possible to add own function there?

parser/parser_test.go
ckganesan: added more tests cases as per your feedback

antonmedv (approved):

antonmedv:
Awesome work!

antonmedv:
I actually not understand what original issues is was about: https://github.com/expr-lang/expr/issues/568

Is to be able to use reuse operator names as functions:

```go
	options := []expr.Option{
		expr.Function(
			"contains",
				return false, nil
			},
			new(func() bool),
		),
	}

	program, err := expr.Compile(`contains()`, options...)
```

I think this is a bad practice and not should be allowed. Also I think it's bad decision to allow to use some operators as functions =/

Let's not processed this this PR @ckganesan?

Q16G:
operator becomes a function to increase the extensibility, convenient for users to customize or extend. Consider a scenario where the user needs to pass in multiple arguments to the contains function, and the original operator is not satisfied, we can extend it to contain from an operator to a function

antonmedv:
> Consider a scenario where the user needs to pass in multiple arguments to the contains function

Can you show some examples?

Q16G:
Will this operation be allowed

antonmedv:
No. For now, not. I don't see a use case. And adding such feature complicates parser and type checker.

PR #579, opened 2024-02-26
add builtin functions : pow and switch

tinboxw:
add builtin functions : pow and switch


antonmedv:
Please, elaborate why do we need pow() builtin? We already have `x**y` and `x^y`.

For switch() builtin we better implement some syntax. Maybe it will be better to do something like `match`.

tinboxw:
Sorry, I overlooked that the `pow(^/**)` has already been implemented while reviewing the documentation. It seems unnecessary to implement `pow` again. 
Regarding the `switch()` functionality, I hope to achieve a more concise and clear syntax to replace the chaining of ternary operators`?:`.
input data like:
`{"distance": 100}`
I hope to calculate a value based on the distance, but it seems that I can only achieve this functionality using the built-in ternary operator, for example:
```
(distance>300 || distance == 0)? -1
  :(distance > 250? 5
   :(distance > 200? 4
    :(distance > 150? 3
     :(distance > 100 ? 2
      :(distance > 50 ? 1
       :(distance > 0? 0)
       )
      )
     )
    )
   )
```
So, I implemented a switch to replace the ternary operator. 
```
switch(distance > 300 || distance == 0, -1, 
       distance > 250, 5, 
       distance > 200, 4, 
       distance > 150, 3, 
       distance > 60,  2, 
       distance > 30,  1, 
       distance > 0,   0)
```
Because I'm not very familiar with the expr project, the implemented `switch()` might not be particularly user-friendly. I hope you can help optimize it.

antonmedv:
image
I see the point! 

But implementing it as function not the best solution. As it will require to evaluate all conditions and cases before choosing a correct one. 

We need to create a proper switch syntax in Expr.

PR #580, opened 2024-02-26
Enable Support for Arrays in Sum, Mean, and Median Functions

ckganesan:
This PR enhances the functionality of the sum, mean, and median functions to support arrays as input values. Now, users can conveniently calculate the sum, mean, and median of arrays of values, expanding the versatility and usability of these functions. 

Implemented based on received feedback. https://github.com/expr-lang/expr/pull/576#issuecomment-1962701702


antonmedv:
Lets not add this feature to `sum`. I'm planning to convert sum to predicate form:

```
Events | sum(.Value)
```

ckganesan:
> Lets not add this feature to `sum`. I'm planning to convert sum to predicate form:
> 
> ```
> Events | sum(.Value)
> ```

I believe this won't impact the functionality of the predicate.

PR #581, opened 2024-02-27
Support chained comparison

bizywizy:
https://github.com/expr-lang/expr/issues/549

### DIFF
--- a/parser/parser_test.go
+++ b/parser/parser_test.go
@@ -531,6 +531,66 @@ world`},
 				To:   &IntegerNode{Value: 3},
 			},
 		},
+		{
+			`1 < 2 > 3`,
+			&BinaryNode{
+				Operator: "&&",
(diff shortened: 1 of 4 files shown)


parser/parser_test.go
```diff
@@ -531,6 +531,46 @@ world`},
...
+						Right:    &IntegerNode{Value: 3},
+					},
+				},
+				Right: &BinaryNode{
+					Operator: "<",
+					Left:     &IntegerNode{Value: 3},
+					Right:    &IntegerNode{Value: 4},
+				},
+			},
+		},
```
antonmedv, line 573: Lets also add `1 < 2 < 3 == true` test case?

bizywizy: Sure, done

antonmedv (approved):

Issue #573, opened 2024-02-20
Deep nil coalescing 


antonmedv:
Right now, in order to use `??` user have to use `?.` instead of `.` in nested structures. 

```js
events[0]?.labels?.value ?? 'defaul'
```

My idea is to make left hand side of `??` optional by default. So in case user written:

```js
events[0].labels.value ?? 'defaul'
```

All member accesses to label/value/events will be optional. As well as array element access. 

I really like this idea and think it makes much sense.

antonmedv:
Also implement opcode to do chained optional fetch

ckganesan:
@antonmedv : I have addressed these issues in the pull request. Could you please review this PR #571

antonmedv:
Cool! Will take a look!

PR #583, opened 2024-02-27
feat: support optional array, string access and deep nil coalescing

ckganesan:
This PR is raised to add support for optional string and array access. [#540](https://github.com/expr-lang/expr/issues/540 ) [#547](https://github.com/expr-lang/expr/issues/547) [#573](https://github.com/expr-lang/expr/issues/573)


expr_test.go
```diff
@@ -1273,6 +1273,18 @@ func TestExpr(t *testing.T) {
...
+		{
+			`[1, 2, 3]?.[5]`,
+			nil,
+		},
+		{
+			`'string'?.[5]`,
+			"g",
+		},
+		{
+			`'string'?.[7]`,
```
antonmedv, line 1285: This is not exactly correct behavior I'm looking for. 

In case `array?.[7]` operator `?.` checks what array itself defined and not nil.

So the code

```
array?.[7]
```
will panic in there is no 7 element index. 

What we can do is to alter `[7]` behavior with operator `?.` **after** it.
 
```
array[7]?.field
```

In this case we can alter OpFetch and do not panic if 7 index does not exist. 

This mean we, unfortunately, does not have a syntax to alter `[7]` behavior if it is the last postfix op.

But we can alter the whole "chain" in presence of `??` operator! 

```js
array[7].field ?? 'default'
```

antonmedv (changes requested):

Issue #589, opened 2024-03-01
Issue with Not Operator Handling in Binary Node Expressions


ckganesan:
This issue pertains to the handling of the "not" operator within binary node expressions. Currently, there seems to be a discrepancy or error in how the "not" operator is processed, leading to unexpected behavior or incorrect results in certain contexts. #590 

Example:

```go
-1 not in [1, 2, 3, 4] // Expectation: true Output: false
1 * 8 not in [1, 2, 3, 4] // Expectation: true Output: false
'foo' + 'bar' not matches 'foobar' // Expectation: false Output: true
```

PR #590, opened 2024-03-01
Resolved 'Not' Operator Issue in Binary Node Expression

ckganesan:
This pull request addresses issue #589.

### DIFF
--- a/compiler/compiler_test.go
+++ b/compiler/compiler_test.go
@@ -551,15 +551,6 @@ func TestCompile_optimizes_jumps(t *testing.T) {
 			},
 		},
 		{
-			`-1 not in [1, 2, 5]`,
-			[]op{
-				{vm.OpPush, 0},
-				{vm.OpPush, 1},
-				{vm.OpIn, 0},
-				{vm.OpNot, 0},
-			},
-		},
-		{
 			`1 + 8 not in [1, 2, 5]`,
 			[]op{
 				{vm.OpPush, 0},
--- a/compiler/compiler_test.go
+++ b/compiler/compiler_test.go
@@ -541,6 +541,48 @@ func TestCompile_optimizes_jumps(t *testing.T) {
 				{vm.OpFetch, 0},
 			},
 		},
+		{
+			`-1 not in [1, 2, 5]`,
+			[]op{
+				{vm.OpPush, 0},
+				{vm.OpPush, 1},
+				{vm.OpIn, 0},
+				{vm.OpNot, 0},
+			},
+		},
+		{
+			`-1 not in [1, 2, 5]`,
+			[]op{
+				{vm.OpPush, 0},
+				{vm.OpPush, 1},
+				{vm.OpIn, 0},
+				{vm.OpNot, 0},
+			},
+		},
+		{
+			`1 + 8 not in [1, 2, 5]`,
+			[]op{
+				{vm.OpPush, 0},
+				{vm.OpPush, 1},
+				{vm.OpIn, 0},
+				{vm.OpNot, 0},
+			},
+		},
+		{
+			`true ? false : 8 not in [1, 2, 5]`,
+			[]op{
+				{vm.OpTrue, 0},
+				{vm.OpJumpIfFalse, 3},
+				{vm.OpPop, 0},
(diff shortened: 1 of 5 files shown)


parser/parser.go
```diff
@@ -126,48 +126,48 @@ func (p *parser) expect(kind Kind, values ...string) {
...
+		return p.parseVariableDeclaration()
 	}
 
 	nodeLeft := p.parsePrimary()
 
 	prevOperator := ""
-	opToken := p.current
-	for opToken.Is(Operator) && p.err == nil {
-		negate := false
+	for opToken := p.current; opToken.Is(Operator) && p.err == nil; opToken = p.current {
```
antonmedv, line 136: I think it's better to keep the old formatting, as we iterating not only on `opToken`, but also updating `prevOperator` on each cycle.

parser/operator/operator.go
```diff
@@ -20,6 +20,15 @@ func IsBoolean(op string) bool {
 	return op == "and" || op == "or" || op == "&&" || op == "||"
 }
 
+func IsFunc(op string) bool {
```
antonmedv, line 23: I know we have a PR with this func in separate PR. But for this PR lets rename the func to something like:  `AllowNegatePrefix`

parser/parser_test.go
```diff
@@ -365,6 +365,62 @@ world`},
...
+					},
+					Right: &ArrayNode{Nodes: []Node{
+						&IntegerNode{Value: 1},
+						&IntegerNode{Value: 2},
+						&IntegerNode{Value: 3},
+						&IntegerNode{Value: 4},
+					}}}},
+		},
+		{
+			"2==2 ? false : 3 not in [1, 2, 5]",
```
antonmedv, line 396: Will be cool to require a brackets `(` `)` in this case too. As even for me, it's unclear what operator will be applied first.

antonmedv:
```
2 == 2 ? false : (3 not in [1, 2, 5])
```

But we can implement it in a new PR.

parser/parser_test.go
```diff
@@ -706,6 +762,11 @@ invalid float literal: strconv.ParseFloat: parsing "0o1E+1": invalid syntax (1:6
 invalid float literal: strconv.ParseFloat: parsing "1E": invalid syntax (1:2)
  | 1E
  | .^
+
+1 not == [1, 2, 5]
+unexpected token Operator("==") (1:7)
```
antonmedv, line 767: Nice! I like this test.

parser/parser.go
```diff
@@ -126,48 +126,48 @@ func (p *parser) expect(kind Kind, values ...string) {
 // parse functions
 
 func (p *parser) parseExpression(precedence int) Node {
-	if precedence == 0 {
-		if p.current.Is(Operator, "let") {
-			return p.parseVariableDeclaration()
-		}
+	if precedence == 0 && p.current.Is(Operator, "let") {
```
antonmedv, line 129: Nice catch 👍🏻

antonmedv (changes requested):

compiler/compiler_test.go
```diff
@@ -541,6 +541,48 @@ func TestCompile_optimizes_jumps(t *testing.T) {
...
+			`-1 not in [1, 2, 5]`,
+			[]op{
+				{vm.OpPush, 0},
+				{vm.OpPush, 1},
+				{vm.OpIn, 0},
+				{vm.OpNot, 0},
+			},
+		},
+		{
+			`-1 not in [1, 2, 5]`,
```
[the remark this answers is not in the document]
ckganesan: Duplicate removed based on your feedback.

parser/parser.go
ckganesan: Reverted to old format as per your feedback.

parser/operator/operator.go
ckganesan: Renamed function name as "AllowedNegateSuffix"

parser/parser_test.go
ckganesan: This behavior aligns with expectations since the ':' operator is not considered a comparison operator.

PR #592, opened 2024-03-04
Add predicate to sum() builtin

antonmedv:


antonmedv:
Will merge after more benchmarking.

Issue #597, opened 2024-03-12
Array function filter causes type information to be lost


SamCymbaluk:
The built in `filter` function takes an array of type `T` and returns a reduced array of type `T`, but the checker currently loses this type information; if the input to `filter` is `[]T`, the output will be `[]any`.

antonmedv:
Fixed in #667!

PR #598, opened 2024-03-12
Prevent filter and array from needlessly weakening types

SamCymbaluk:
fixes: https://github.com/expr-lang/expr/issues/597


antonmedv:
This is done in #667!

PR #614, opened 2024-03-27
feat: extract code for compiling equal operator

zhuliquan:
Whether `OpEqualInt` and `OpEqualInt` need to be applied to the operator `!=` when convert ast to bytecode, and if so, whether to consider extracting the code that handles equal into a function named `equalBinnaryNode`

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -395,34 +395,12 @@ func (c *compiler) UnaryNode(node *ast.UnaryNode) {
 }
 
 func (c *compiler) BinaryNode(node *ast.BinaryNode) {
-	l := kind(node.Left)
-	r := kind(node.Right)
-
-	leftIsSimple := isSimpleType(node.Left)
-	rightIsSimple := isSimpleType(node.Right)
-	leftAndRightAreSimple := leftIsSimple && rightIsSimple
-
 	switch node.Operator {
 	case "==":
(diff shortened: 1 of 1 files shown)


antonmedv:
Why this code is needed?

zhuliquan:
> Why this code is needed?

I'm not sure. I notice that you have applied three `OpCode` in equal operator case as below.
https://github.com/expr-lang/expr/blob/d5231073da6742e42f5c026eb49c0d493da1274c/compiler/compiler.go#L412-L418
I'm guessing you want to make the most of the right and left type information in `BinaryNode`, so that the `Bytecode` generated later runs faster. Because. I notice that assert `interface{}` to `int` / `string` directly in case `OpEqualInt`/`OpEqualString` as below. 
https://github.com/expr-lang/expr/blob/d5231073da6742e42f5c026eb49c0d493da1274c/vm/vm.go#L165-L173
Since this trick is already used in case of `==`, why not use it in case of `!=`? Further, I think that if this trick can be used in case of `!=`, why not extract code of `==` case and apply it to both the case of `==` and the case of `!=`.

antonmedv (approved):

antonmedv:
> Since this trick is already used in case of ==, why not use it in case of !=?

Got it. This makes sense.

Issue #613, opened 2024-03-26
Parentheses disappear for conditional node


aizeNR:
Hi, when you build a tree with a conditional expression, and convert it to a tree and back to an expression, the parentheses for the conditional disappear.

Below is code to reproduce
https://go.dev/play/p/mHo7-d3vFuo

```
package main

import (
	"fmt"

	"github.com/expr-lang/expr/parser"
)

func main() {
	tree, err := parser.Parse("(UserID != 0 ? true : false) && ProductID == 1")
	if err != nil {
		fmt.Print(err.Error())
		return
	}

	fmt.Print(tree.Node.String())
}
```

zhuliquan:
may be below branch will help you
https://github.com/zhuliquan/expr/tree/feature-parenthesis-node

antonmedv:
Fixed.

PR #620, opened 2024-04-06
Feature parenthesis node

zhuliquan:
I found there are many operator of adding parentheses in ast/print.go
for example:
https://github.com/expr-lang/expr/blob/7772ea0fee5c9b153439a0ad5ee2e2b1a764f453/ast/print.go#L54-L56
https://github.com/expr-lang/expr/blob/7772ea0fee5c9b153439a0ad5ee2e2b1a764f453/ast/print.go#L91-L101
why not add a specific type ast node for following benefits:
1、make printing string more like the expr inputted by user, i.e. resolve issue: https://github.com/expr-lang/expr/issues/613.
2、Make the node's responsibilities more single.

### DIFF
--- a/ast/node.go
+++ b/ast/node.go
@@ -214,3 +214,12 @@ type PairNode struct {
 	Key   Node // Key of the pair.
 	Value Node // Value of the pair.
 }
+
+// ParenthesisNode represents a node with parenthesis.
+// Example:
+//
+//	(foo and bar)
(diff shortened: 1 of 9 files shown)


antonmedv:
This is a interesting approach, unfortunately it makes other things more complex. Instead we should just feap thenary nodes in parenthesis if inside binary.

PR #621, opened 2024-04-07
Feature parenthesis prop

zhuliquan:
Add `hasParen` field for ast.Ndoe, which indicated add parenthesis when invoking `String()`. 
I've always thought that you should keep the parenthetical information of the node if the user input it.

### DIFF
--- a/ast/node.go
+++ b/ast/node.go
@@ -13,6 +13,7 @@ type Node interface {
 	Type() reflect.Type
 	SetType(reflect.Type)
 	String() string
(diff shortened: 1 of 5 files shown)


antonmedv:
Although I think this is a cool idea, it shouldn't be a part of AST. And default print node to string should not rely on it. 

Usually in other programming languages this is done by additional linters with little bit more sophisticated parsers which also keeps information about new lines, brackets, types of quotes, comments. 

Creating such linter for expr will the super nice.

PR #623, opened 2024-04-09
Allow tailing comma in arguments

zhuliquan:
allow the last argument append with comma in func expr.
make below expr can be parsed ok.
```expr
func1(
     argument1,
     argument2,
     argument3,
)
```


antonmedv:
Althoug this is nice feature, I think for not is not needed in Expr. as most of the time expressions is one line.

We may return later to this and change our decision.

antonmedv:
Hmm. This is true. We do allow comma in arrays & maps. Let me think on it bit more.

antonmedv:
Well, let's add support for this!

PR #633, opened 2024-04-23
Preserve The Specified Timezone In Parse

cwarden:
When parsing a string into a Time using the optional timezone parameter, set the location of the Time to the timezone's location so that Format uses that timezone.

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -502,6 +502,7 @@ var Builtins = []*Function{
 				if err != nil {
 					return nil, err
 				}
+				t = t.In(tz)
 				return t, nil
 			}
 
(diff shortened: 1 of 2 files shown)


antonmedv:
Let's add mode tests for this use case?

antonmedv (approved):

antonmedv:
I think we not need this PR any more.

Here is the logic:

If parsed date has a TZ data `+0100`, ParseInLocation uses it.

```go
date("2023-04-23T00:30:00.000+0100", "2006-01-02T15:04:05-0700", "America/Chicago")
2023-04-23 00:30:00 +0100 +0100
```

Without TZ data, third param is used as TZ:

```go
date("2023-04-23T00:30:00.000", "2006-01-02T15:04:05", "America/Chicago")
2023-04-23 00:30:00 -0500 CDT
```

So we do not need implicit convert with In any more.

antonmedv:
```go
❯ date("2023-04-23T00:30:00.000", "2006-01-02T15:04:05", "America/Chicago")
2023-04-23 00:30:00 -0500 CDT
❯ date("2023-04-23T00:30:00.000", "2006-01-02T15:04:05", "America/Chicago").In(timezone("America/Chicago"))
2023-04-23 00:30:00 -0500 CDT
```

Issue #618, opened 2024-04-02
EOL for pmezard/go-difflib dependency


anurag-deshpande:
[Problem Description]
We are consuming `github.com/expr-lang/expr v1.15.7`, which is latest as of date. This package is scanned for security vulnerabilities and EOLs by **blackduck scanner** at our source.

The blackduck scanner has identified a Project EOL component, `pmezard-go-difflib20190219-snapshot-5d4384ee` , which is a **transitive dependency** of `github.com/expr-lang/expr v1.15.7`. This project is not maintained and thus EOLed much earlier.

[Request]
We wish to consume all the dependencies which are non-EOLed, to maintain good coding practices. Can this EOLed component be updated by expr contributors or replaced with some alternative with similar functionality, to reduce the EOL risk?

Let me know if any more information is needed for this issue.

antonmedv:
Expr uses testify (tests only), and testify uses go-difflib. 

Upstream issue: https://github.com/stretchr/testify/issues/1327

Will see what I can do about it.

antonmedv:
Upstream issue is stale. Will vendor testify.

PR #635, opened 2024-04-26
Add Optional Timezone to now()

cwarden:
Add support for an optional timezone argument to now().  Formatting the returned time will use the specified timezone.


go.mod
```diff
@@ -2,7 +2,10 @@ module github.com/expr-lang/expr
 
 go 1.18
 
-require github.com/stretchr/testify v1.8.4
+require (
+	github.com/agiledragon/gomonkey/v2 v2.11.0
```
antonmedv, line 6: Let's not add additional dependencies.

antonmedv (changes requested):

go.mod
cwarden: It's just a test dependency to address the race in the tests.  Any other preferred method for avoiding the race condition, or just leave it?

antonmedv: It is a security concern. I'm also will be happy to avoid testify if possible. I understand what this is test dep, but apparently vet tools not: https://github.com/expr-lang/expr/issues/618

antonmedv:
I've implemented default timezone and `timezone()` builtins.

```go
program, err := expr.Compile(`now().Location().String()`, expr.Timezone("Asia/Kamchatka"))
if err != nil {
	fmt.Printf("%v", err)
	return
}

output, err := expr.Run(program, nil)
if err != nil {
	fmt.Printf("%v", err)
	return
}

fmt.Printf("%v", output)
// Output: Asia/Kamchatka
```

Issue #639, opened 2024-05-06
Double escaping pair node key


aizeNR:
Hi, in [commit](https://github.com/expr-lang/expr/commit/d5231073da6742e42f5c026eb49c0d493da1274c) add escaping pair node key, but str.String() already uses this and it causes double escaping. 

Example https://go.dev/play/p/WTkUxu5yM4e
```go
package main

import (
	"fmt"

	"github.com/expr-lang/expr/parser"
)

func main() {
	tree, err := parser.Parse(`{"9": 10}`)
	if err != nil {
		fmt.Print(err.Error())
		return
	}

	fmt.Print(tree.Node.String())
}
```

PR #640, opened 2024-05-07
Resolve double-escaping issue with pair node key

ckganesan:
This pull request addresses the issue of double escaping in the pair node key. #639


antonmedv:
Let's add a test case.

antonmedv (approved):

PR #678, opened 2024-06-26
Fix print with precedence

zhuliquan:
consider the precedence of operators and their associative law when printing unary and binary nodes
fix error expr print:
1、`(-(1+1)) ** 2` => `-(1 + 1) ** 2`
2、`(2 ** 2) ** 3` => `2 ** 2 ** 3`
3、`(3 + 5) / (5 % 3)` => `(3 + 5) / 5 % 3`

Beside，I think should remove below code:
https://github.com/expr-lang/expr/blob/971388017260b2241784c4ed216f8e6eb91ef7f5/ast/print.go#L75-L80
https://github.com/expr-lang/expr/blob/971388017260b2241784c4ed216f8e6eb91ef7f5/ast/print.go#L86-L88
because, these code logical can be covered with the logical of precedence and associative.


antonmedv:
Super! Thanks a lot.

PR #693, opened 2024-08-02
fix as result not working with int32,int16,int8

### LINKED ISSUE #689: expr.AsKind not work?
```golang
func TestAsKind(t *testing.T) {
	expectValue := int32(0)
	str := "1"
	program, err := expr.Compile(str, expr.AsKind(reflect.TypeOf(expectValue).Kind()))
	if err != nil {
		fmt.Printf("Compile Error: %s, %v\n", str, err)
		return
	}
	v, err := expr.Run(program, nil)
	if err != nil {
		fmt.Printf("Run Error: %s, %v\n", str, err)
		return
	}
	fmt.Println(v)
}
// output:
// Compile Error: 1, expected int32, but got int
```

is it a bug ?

wodeqiangne:
fix #689 for as result not working with int32 int16 and int8

### DIFF
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -79,7 +79,7 @@ func Check(tree *parser.Tree, config *conf.Config) (reflect.Type, error) {
 		}
 
 		switch v.config.Expect {
-		case reflect.Int, reflect.Int64, reflect.Float64:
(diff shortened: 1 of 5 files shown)


antonmedv:
Could you please rebase your branch? Checker got changes a little bit.

PR #700, opened 2024-08-23
Add argo-workflows as a user

Joibel:
Argo-workflows makes use of expr-lang, and this list is both companies and open source products.


antonmedv:
Thanks for your contribution. But we already have Agro. See a few lines before

Joibel:
Argh, I searched badly. Sorry

PR #751, opened 2025-01-25
fix(builtin): handle string pointer in fromJSON

### LINKED ISSUE #739: Can we make the `fromJSON` function support string pointer?
The current `fromJSON` function only supports parameters of type string. If I pass a parameter of type *string, it throws an error.

thevilledev:
Support `*string` input in the `fromJSON` builtin function by runtime type handling. Type switch how to handle each case.

Added tests covering both direct string and pointer-to-string use cases, including `toJSON` function.

Fixes #739

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -443,7 +443,11 @@ var Builtins = []*Function{
 		Name: "fromJSON",
 		Func: func(args ...any) (any, error) {
 			var v any
-			err := json.Unmarshal([]byte(args[0].(string)), &v)
+			jsonStr := args[0]
+			if strPtr, ok := jsonStr.(*string); ok {
+				jsonStr = *strPtr
+			}
+			err := json.Unmarshal([]byte(jsonStr.(string)), &v)
 			if err != nil {
 				return nil, err
 			}
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -463,7 +463,7 @@ var Builtins = []*Function{
 			}
 			return v, nil
 		},
-		Types: types(new(func(string) any), new(func(*string) any)),
+		Types: types(new(func(string) any)),
 	},
 	{
 		Name: "toBase64",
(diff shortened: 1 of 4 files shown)


builtin/builtin.go
```diff
@@ -443,7 +443,11 @@ var Builtins = []*Function{
 		Name: "fromJSON",
 		Func: func(args ...any) (any, error) {
 			var v any
-			err := json.Unmarshal([]byte(args[0].(string)), &v)
+			jsonStr := args[0]
+			if strPtr, ok := jsonStr.(*string); ok {
```
antonmedv, line 447: Althoug this is a working solution, we need to understand why OpDeref was not added before calling fromJSON. 

Compiler adds OpDeref for arguments before calling a function. But only if needed. Aperently the check is not perfect.

antonmedv:
We can merge this PR, and it will fix problem with *string and fromJSON. But we still can encounter those errors.

thevilledev:
I did a quick glance over the compiler earlier but maybe I missed something. Agree that this would ideally be fixed for other use cases as well than just JSON. I can take a look and keep this as a draft PR until then.

thevilledev:
I think I found a way to do this at compile-time, with just a little type hinting at runtime in `fromJSON` builtin. Could be extended to other functions too. Let me know what you think :)

thevilledev:
Well shoot. Looks like my assumption was wrong. Back to the drawing board 😅

thevilledev:
Used this scenario when debugging:

- `&jsonString` is stored in the env map `map[string]any` as `interface{}`
- When `fromJSON(json)` is called, the value is mapped to `interface{}`, not as `*string`
- No OpDeref is triggered
- We get `args[0]` as `interface{}` in the builtin func
- Only runtime fixes available, as visible in this PR

So the compiler can't see through the `interface{}` to know deref is required. During type checking/compilation, we could look up the actual value, use reflection for the real type and insert OpDeref

builtin/builtin.go
antonmedv: I figured out why. Builtins compilation just do not check for deref. So we ether make compiler more advance. Or make it via additional call.

antonmedv:
It was quite challenging, but I was able to fix it: 80f0ea62ea6935a06d6562c5844a801362bf57a1

PR #764, opened 2025-03-09
feat: cpu profile

zhuliquan:
This is proof of concept PR, which help to analyze expr profile. I notice that expr has profile feature, but `Config` doesn't explode it for user (i.e. user can't set `Profile` true).
https://github.com/expr-lang/expr/blob/80f0ea62ea6935a06d6562c5844a801362bf57a1/compiler/compiler.go#L212-L232
I think it's very likely that the reason lies in how you show the profile results. I've noticed that the results are saved to a structure called Span. 
https://github.com/expr-lang/expr/blob/80f0ea62ea6935a06d6562c5844a801362bf57a1/vm/utils.go#L27-L33
My idea is to convert the information stored in the Span result into a pprof file, so that we can use `go tool pprof` command to visualise how the expression is running and find out where it is time-consuming. 

I create new sub-module `profile`. I write a function `GeneratePprofProfile` to convert `Span` struct to pprof file (Of course, this function is not perfect, and the profile may need to be more refined. Profile construct). Finally I write a example code and got below result. 
Screenshot_2025-03-09_17-04-28
Screenshot_2025-03-09_17-06-24
Screenshot_2025-03-09_17-06-38


antonmedv:
This is a nice idea to use pprof. Only for complex rules only reporting ast node type maybe not enough. And actually expr profile is a part of Expr Pro https://expr-lang.org/docs/expr-pro/profile I sell Expr Pro to boost Expr development ;) 
Additionally I created a separate renderer for profile, with information about which part of expression is causing errors:  
image

PR #806, opened 2025-06-24
Allow ignoring struct fields with struct tag expr:"-"

### LINKED ISSUE #805: Feature request: allow ignoring struct fields with struct tag expr:"-"
**Description**: Similar to what other Go libraries provide, it would be nice that in expr we could ignore a struct field in the environment if the field has a struct tag `expr:"-"`.

**Example happening in other libraries**: in the standard library the package `encoding/json` ignores struct fields that have the struct tag `json:"-"`, causing it to neither encode nor decode those fields.

**Use case**: reusing the same struct to avoid tedious copying that is error-prone.

**Example of proposed feature**:
```go
type MyEnv struct {
	VisibleField string
	HiddenField  string `expr:"-"`
}

func main() {
	_, err := expr.Eval(`HiddenField`, MyEnv{})
	// we would get a `cannot fetch HiddenField` error
}
```

**P.S.**: I have created PR https://github.com/expr-lang/expr/pull/806 as an example of how the feature could be implemented. Please, let me know if you would rather discuss a different approach if you think this proposal makes sense.

Thank you :)

diegommm:
Fixes #805

### DIFF
--- a/builtin/lib.go
+++ b/builtin/lib.go
@@ -421,10 +421,11 @@ func get(params ...any) (out any, err error) {
 		fieldName := i.(string)
 		value := v.FieldByNameFunc(func(name string) bool {
 			field, _ := v.Type().FieldByName(name)
-			if field.Tag.Get("expr") == fieldName {
+			tagName := field.Tag.Get("expr")
+			if tagName == fieldName {
 				return true
 			}
-			return name == fieldName
+			return tagName != "-" && name == fieldName
 		})
 		if value.IsValid() {
 			return value.Interface(), nil
--- a/checker/nature/utils.go
+++ b/checker/nature/utils.go
@@ -7,10 +7,14 @@ import (
 )
 
 func fieldName(field reflect.StructField) string {
-	if taggedName := field.Tag.Get("expr"); taggedName != "" {
+	switch taggedName := field.Tag.Get("expr"); taggedName {
+	case "-":
+		return ""
+	case "":
+		return field.Name
+	default:
 		return taggedName
 	}
-	return field.Name
 }
 
 func fetchField(t reflect.Type, name string) (reflect.StructField, bool) {
@@ -23,7 +27,7 @@ func fetchField(t reflect.Type, name string) (reflect.StructField, bool) {
 	for i := 0; i < t.NumField(); i++ {
 		field := t.Field(i)
 		// Search all fields, even embedded structs.
-		if fieldName(field) == name {
+		if n := fieldName(field); n != "" && n == name {
 			return field, true
 		}
 	}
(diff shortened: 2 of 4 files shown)


expr_test.go
```diff
@@ -139,7 +140,71 @@ func ExampleEnv_tagged_field_names() {
 
 	fmt.Printf("%v", output)
 
-	// Output : Hello World
```
diegommm, line 142: This is unrelated but I realized that this example test was not being run because there is a space between `Output` and `:`, so I removed the space and the example test is now running and is successful.

builtin/lib.go
```diff
@@ -421,10 +421,11 @@ func get(params ...any) (out any, err error) {
 		fieldName := i.(string)
 		value := v.FieldByNameFunc(func(name string) bool {
 			field, _ := v.Type().FieldByName(name)
-			if field.Tag.Get("expr") == fieldName {
+			tagName := field.Tag.Get("expr")
+			if tagName == fieldName {
 				return true
 			}
-			return name == fieldName
+			return tagName != "-" && name == fieldName
```
antonmedv, line 428: Lets simplify logic here. Use early exit:
```go
if tagName == "-" {
```

checker/nature/utils.go
```diff
@@ -23,7 +27,7 @@ func fetchField(t reflect.Type, name string) (reflect.StructField, bool) {
 	for i := 0; i < t.NumField(); i++ {
 		field := t.Field(i)
 		// Search all fields, even embedded structs.
-		if fieldName(field) == name {
+		if n := fieldName(field); n != "" && n == name {
```
antonmedv, line 30: Lets return str, bool and explicitly check if field is allowed.

checker/nature/utils.go
```diff
@@ -7,10 +7,14 @@ import (
 )
 
 func fieldName(field reflect.StructField) string {
-	if taggedName := field.Tag.Get("expr"); taggedName != "" {
+	switch taggedName := field.Tag.Get("expr"); taggedName {
+	case "-":
+		return ""
```
antonmedv, line 12: Better to return 
```go
return "", false
```

checker/nature/utils.go
```diff
@@ -69,7 +73,11 @@ func StructFields(t reflect.Type) map[string]Nature {
 				}
 			}
 
-			table[fieldName(f)] = Nature{
+			name := fieldName(f)
+			if name == "" {
```
antonmedv, line 77: Same, here.

vm/runtime/runtime.go
```diff
@@ -65,10 +65,11 @@ func Fetch(from, i any) any {
 		fieldName := i.(string)
 		value := v.FieldByNameFunc(func(name string) bool {
 			field, _ := v.Type().FieldByName(name)
-			if field.Tag.Get("expr") == fieldName {
+			tagName := field.Tag.Get("expr")
+			if tagName == fieldName {
 				return true
 			}
-			return name == fieldName
+			return tagName != "-" && name == fieldName
```
antonmedv, line 72: Same here.

PR #811, opened 2025-06-28
Reduce lexer memory allocs

### LINKED ISSUE #810: Reduce allocations during parsing
Current parsing (and lexing) is very efficient in terms of direct CPU and memory usage, but it has an indirect cost on CPU because it performs many little allocations, which increases GC pressure.
In use cases where a large number of expressions need to be parsed, this can cause spikes in allocations whereas it can be rather easy to avoid.

I will provide a PR with improvements and benchmarks.

diegommm:
Fixes #810 

Benchmark results (20% faster, 40% less memory copying, 75% less allocations):
```
goos: darwin
goarch: arm64
pkg: github.com/expr-lang/expr/parser/lexer
cpu: Apple M4 Pro
          │   old.txt   │               new.txt               │
          │   sec/op    │   sec/op     vs base                │
Parser-12   1.549µ ± 1%   1.254µ ± 1%  -19.05% (p=0.000 n=20)

          │   old.txt    │               new.txt                │
Parser-12   2.961Ki ± 0%   1.789Ki ± 0%  -39.58% (p=0.000 n=20)

          │   old.txt   │              new.txt               │
          │  allocs/op  │ allocs/op   vs base                │
Parser-12   27.000 ± 0%   7.000 ± 0%  -74.07% (p=0.000 n=20)
```

Benchmarks were performed in the `parser/lexer` directory executing the command:
```
go test -run=zzz-no-tests -bench=. -count=20 > file
```
The file was renamed to either `old.txt` or `new.txt`, and the comparison was made with:
```
benchstat old.txt new.txt
```

<details><summary>Raw results from old.txt</summary>
<p>

```
goos: darwin
goarch: arm64
pkg: github.com/expr-lang/expr/parser/lexer
cpu: Apple M4 Pro
BenchmarkParser-12        790826              1559 ns/op            3032 B/op         27 allocs/op
BenchmarkParser-12        727635              1547 ns/op            3032 B/op         27 allocs/op
BenchmarkParser-12        657169              1522 ns/op            3032 B/op         27 allocs/op
BenchmarkParser-12        670780              1572 ns/op            3032 B/op         27 allocs/op
BenchmarkParser-12        739662              1558 ns/op            3032 B/op         27 allocs/op
BenchmarkParser-12        763692              1560 ns/op            3032 B/op         27 allocs/op
BenchmarkParser-12        749037              1550 ns/op            3032 B/op         27 allocs/op
BenchmarkParser-12        761952              1526 ns/op            3032 B/op         27 allocs/op
BenchmarkParser-12        791458              1524 ns/op            3032 B/op         27 allocs/op
BenchmarkParser-12        757533              1533 ns/op            3032 B/op         27 allocs/op
BenchmarkParser-12        762931              1606 ns/op            3032 B/op         27 allocs/op
BenchmarkParser-12        752461              1593 ns/op            3032 B/op         27 allocs/op
BenchmarkParser-12        727683              1545 ns/op            3032 B/op         27 allocs/op
BenchmarkParser-12        775984              1534 ns/op            3032 B/op         27 allocs/op
BenchmarkParser-12        771271              1534 ns/op            3032 B/op         27 allocs/op
BenchmarkParser-12        783654              1582 ns/op            3032 B/op         27 allocs/op
BenchmarkParser-12        774580              1547 ns/op            3032 B/op         27 allocs/op
BenchmarkParser-12        754653              1539 ns/op            3032 B/op         27 allocs/op
BenchmarkParser-12        758157              1568 ns/op            3032 B/op         27 allocs/op
BenchmarkParser-12        754888              1551 ns/op            3032 B/op         27 allocs/op
PASS
ok      github.com/expr-lang/expr/parser/lexer  24.998s
```

</p>
</details> 

<details><summary>Raw results from new.txt</summary>
<p>

```
goos: darwin
goarch: arm64
pkg: github.com/expr-lang/expr/parser/lexer
cpu: Apple M4 Pro
BenchmarkParser-12        883687              1250 ns/op            1832 B/op          7 allocs/op
BenchmarkParser-12        926926              1229 ns/op            1832 B/op          7 allocs/op
BenchmarkParser-12        965379              1227 ns/op            1832 B/op          7 allocs/op
BenchmarkParser-12        961345              1251 ns/op            1832 B/op          7 allocs/op
BenchmarkParser-12        902880              1332 ns/op            1832 B/op          7 allocs/op
BenchmarkParser-12        880496              1368 ns/op            1832 B/op          7 allocs/op
BenchmarkParser-12        958262              1236 ns/op            1832 B/op          7 allocs/op
BenchmarkParser-12        933210              1241 ns/op            1832 B/op          7 allocs/op
BenchmarkParser-12        941917              1271 ns/op            1832 B/op          7 allocs/op
BenchmarkParser-12        886873              1263 ns/op            1832 B/op          7 allocs/op
BenchmarkParser-12        927643              1251 ns/op            1832 B/op          7 allocs/op
BenchmarkParser-12        960315              1257 ns/op            1832 B/op          7 allocs/op
BenchmarkParser-12        944788              1263 ns/op            1832 B/op          7 allocs/op
BenchmarkParser-12        904806              1267 ns/op            1832 B/op          7 allocs/op
BenchmarkParser-12        938611              1243 ns/op            1832 B/op          7 allocs/op
BenchmarkParser-12        947118              1256 ns/op            1832 B/op          7 allocs/op
BenchmarkParser-12        946309              1251 ns/op            1832 B/op          7 allocs/op
BenchmarkParser-12        948300              1293 ns/op            1832 B/op          7 allocs/op
BenchmarkParser-12        909182              1260 ns/op            1832 B/op          7 allocs/op
BenchmarkParser-12        933751              1251 ns/op            1832 B/op          7 allocs/op
PASS
ok      github.com/expr-lang/expr/parser/lexer  25.899s
```

</p>
</details>

### DIFF
--- a/file/source.go
+++ b/file/source.go
@@ -1,48 +1,36 @@
 package file
 
-import (
-	"strings"
-	"unicode/utf8"
-)
+import "strings"
 
-type Source []rune
+type Source struct {
+	raw string
+}
 
 func NewSource(contents string) Source {
-	return []rune(contents)
+	return Source{
+		raw: contents,
+	}
 }
 
 func (s Source) String() string {
-	return string(s)
+	return s.raw
 }
 
 func (s Source) Snippet(line int) (string, bool) {
-	if s == nil {
+	if s.raw == "" {
 		return "", false
 	}
-	lines := strings.Split(string(s), "\n")
-	lineOffsets := make([]int, len(lines))
-	var offset int
-	for i, line := range lines {
-		offset = offset + utf8.RuneCountInString(line) + 1
-		lineOffsets[i] = offset
+	var start int
+	for i := 1; i < line; i++ {
+		pos := strings.IndexByte(s.raw[start:], '\n')
+		if pos < 0 {
+			return "", false
+		}
+		start += pos + 1
 	}
-	charStart, found := getLineOffset(lineOffsets, line)
-	if !found || len(s) == 0 {
-		return "", false
+	end := start + strings.IndexByte(s.raw[start:], '\n')
+	if end < start {
+		end = len(s.raw)
 	}
-	charEnd, found := getLineOffset(lineOffsets, line+1)
-	if found {
-		return string(s[charStart : charEnd-1]), true
-	}
-	return string(s[charStart:]), true
-}
-
-func getLineOffset(lineOffsets []int, line int) (int, bool) {
-	if line == 1 {
-		return 0, true
-	} else if line > 1 && line <= len(lineOffsets) {
-		offset := lineOffsets[line-2]
-		return offset, true
-	}
-	return -1, false
+	return s.raw[start:end], true
 }
--- a/parser/lexer/lexer.go
+++ b/parser/lexer/lexer.go
@@ -9,7 +9,7 @@ import (
 
 func Lex(source file.Source) ([]Token, error) {
 	l := &lexer{
-		source: source,
+		source: []rune(source.String()),
 		tokens: make([]Token, 0),
 		start:  0,
 		end:    0,
@@ -28,7 +28,7 @@ func Lex(source file.Source) ([]Token, error) {
 }
 
 type lexer struct {
-	source     file.Source
+	source     []rune
 	tokens     []Token
 	start, end int
 	err        *file.Error
--- a/parser/lexer/lexer.go
+++ b/parser/lexer/lexer.go
@@ -8,13 +8,12 @@ import (
 )
 
 func Lex(source file.Source) ([]Token, error) {
+	raw := source.String()
 	l := &lexer{
-		source: []rune(source.String()),
+		raw:    raw,
+		runes:  []rune(raw),
 		tokens: make([]Token, 0),
-		start:  0,
-		end:    0,
 	}
-	l.commit()
 
 	for state := root; state != nil; {
 		state = state(l)
@@ -28,10 +27,16 @@ func Lex(source file.Source) ([]Token, error) {
 }
 
 type lexer struct {
-	source     []rune
+	raw        string
+	runes      []rune
 	tokens     []Token
-	start, end int
 	err        *file.Error
+	start, end pos
+	eof        bool
+}
+
+type pos struct {
+	byte, rune int
 }
 
 const eof rune = -1
@@ -41,12 +46,12 @@ func (l *lexer) commit() {
 }
 
 func (l *lexer) next() rune {
-	if l.end >= len(l.source) {
-		l.end++
+	if l.end.rune >= len(l.runes) {
+		l.eof = true
 		return eof
 	}
-	r := l.source[l.end]
-	l.end++
+	r := l.runes[l.end.rune]
+	l.end.rune++
 	return r
 }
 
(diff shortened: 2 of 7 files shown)


file/source.go
```diff
@@ -1,48 +1,36 @@
...
+		raw: contents,
+	}
 }
 
 func (s Source) String() string {
-	return string(s)
+	return s.raw
 }
 
 func (s Source) Snippet(line int) (string, bool) {
```
diegommm, line 19: This method is no longer used but I'm keeping it in case someone was using it.

parser/lexer/lexer.go
```diff
@@ -3,18 +3,19 @@ package lexer
 import (
 	"fmt"
 	"strings"
+	"unicode/utf8"
 
 	"github.com/expr-lang/expr/file"
 )
 
+const minTokens = 10
```
diegommm, line 11: This allows for optimistically allocating tokens. Discussing the "right" value for this parameter can be a loss of time (everyone has an opinion on what should be the "right value).
The only thing I will hold is that using zero (the default) is the worst of all values because you will likely always need some nodes. It would be very rare that you don't. And the first few times your add a new node with `append` then the runtime will just have to copy over and over the same underlying array while it grows the slice.

parser/lexer/lexer_test.go
```diff
@@ -335,6 +335,7 @@ literal not terminated (1:10)
 früh ♥︎
 unrecognized character: U+2665 '♥' (1:6)
  | früh ♥︎
+ | .....^
```
diegommm, line 338: This appeared to be a small bug. In the original code, it found a rune longer than 1 byte it would skip the indicator line entirely. I imagine that wouldn't be good if I have something like:
```
let myStr = "Hello, 世界"; someError
```
Then it wouldn't put the indicator line because it will find the '世' rune.

file/source.go
```diff
@@ -1,48 +1,36 @@
 package file
 
-import (
-	"strings"
-	"unicode/utf8"
-)
+import "strings"
 
-type Source []rune
+type Source struct {
```
antonmedv, line 5: Maybe just abandon Source altogether? And simply use string?

antonmedv (approved):

parser/lexer/lexer_test.go
antonmedv: Nice, I guess the problem is this rune width in terminal.

diegommm:
@antonmedv I have just committed new changes that allow the parser to use the new iterator API and made new benchmarks on the overall process of parsing (with iterator) instead of benchmarking only the lexer.

Checkout the new results! It's a lot. I'm done for this PR.

file/source.go
diegommm: Agree, I can do that really quick or in a separate PR if you prefer for easier reviewing.

antonmedv:
I just realized: why ring buffer needed at all?

diegommm:
> I just realized: why ring buffer needed at all?

I tried a simpler approach with just a slice of `Token` but it quickly grew uglier and harder to read.

PR #813, opened 2025-07-04
Fix *vm.VM memory leak

diegommm:
Currently, it's reslicing using `[0:0]` (which is equivalent to `[:0]`), but this does not clear the memory references in the underlying array, which prevents garbage collection. For instance, if:

- A VM runs a program and pushes N items to the stack
- The last item is (or contains) a pointer or reference type
- The VM is later reused, but it always pushes less than N items to the stack

Then the Nth old item will not be garbage collected because there is still a reference to it in the underlying array.

So we need to "clear the tail" of the slice before reslicing to avoid this. See these example issues:

- https://go.dev/blog/slices-intro#a-possible-gotcha
- https://go.dev/blog/generic-slice-functions#the-fix

### DIFF
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -67,9 +67,11 @@ func (vm *VM) Run(program *Program, env any) (_ any, err error) {
 	if vm.Stack == nil {
 		vm.Stack = make([]any, 0, 2)
 	} else {
+		clearSlice(vm.Stack)
 		vm.Stack = vm.Stack[0:0]
 	}
 	if vm.Scopes != nil {
+		clearSlice(vm.Scopes)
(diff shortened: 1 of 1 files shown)


antonmedv:
Slightly slower bench, but this is expected.
```
_expr_reuseVm-4               70.73n ± 1%   75.88n ± 1%   +7.28% (p=0.001 n=10)
```

Reusing VM still faster than creating stack vm each time:
```
_expr-4                       114.2n ± 0%   119.9n ± 3%   +4.99% (p=0.000 n=10)
_expr_reuseVm-4               70.73n ± 1%   75.88n ± 1%   +7.28% (p=0.001 n=10)
```

PR #824, opened 2025-08-29
[WIP] Improve Performance of the Type System

diegommm:


antonmedv:
Awesome, it's really interesting to see the results. I think this will be very beneficial for users who use Eval path and do not precompile expressions.

checker/nature/nature.go
```diff
@@ -1,261 +1,599 @@
...
-	Type            reflect.Type      // Type of the value. If nil, then value is unknown.
-	Func            *builtin.Function // Used to pass function type from callee to CallNode.
-	ArrayOf         *Nature           // Elem nature of array type (usually Type is []any, but ArrayOf can be any nature).
-	PredicateOut    *Nature           // Out nature of predicate.
+	// The order of the fields matter, check alignment before making changes.
+
+	Type reflect.Type // Type of the value. If nil, then value is unknown.
+	Kind reflect.Kind // Kind of the value.
+
+	cache *Cache
```
antonmedv, line 40: As this cache is shared, let's do not set it inside the nature itself, but pass it from the top, from the checker as function arguments.

We already have this problem and we have to set a nature cache, for example, in the following line:
```go
		base := n.Node.Nature()
		base.SetCache(c) // AST doesn't cache nature info
		base = base.Deref()
```

Let's not set cash inside an agent. But pass in to all functions what may use it:

```go
		base := n.Node.Nature()
		base = base.Deref(c) // As the deref can modify the cache.
```

conf/env.go
```diff
@@ -9,35 +9,36 @@ import (
 	"github.com/expr-lang/expr/types"
 )
 
-func Env(env any) Nature {
+func Env(c *Cache, env any) Nature {
```
antonmedv, line 12: Lets not change the conf.Env API, I believe it may be used externally in some degree.

conf/config.go
```diff
@@ -34,6 +34,7 @@ type Config struct {
 	Functions FunctionsTable
 	Builtins  FunctionsTable
 	Disabled  map[string]bool // disabled builtins
+	NtCache   nature.Cache
```
antonmedv, line 37: Same here, if we can pass it via the function inside the compiler, from the top it will be easier and we don't need to set it here.

antonmedv (changes requested):

antonmedv:
```
benchstat old.txt new.txt 
goos: darwin
goarch: arm64
pkg: github.com/expr-lang/expr/checker
cpu: Apple M2
                                           │   old.txt    │               new.txt               │
                                           │    sec/op    │   sec/op     vs base                │
Checker/name=function_calls-8                18.788µ ± 3%   7.443µ ± 0%  -60.38% (p=0.000 n=10)
Checker/name=unary_and_binary_operations-8   10.010µ ± 0%   7.997µ ± 2%  -20.11% (p=0.000 n=10)
Checker/name=deep_struct_access-8             30.04µ ± 0%   21.69µ ± 5%  -27.80% (p=0.000 n=10)
geomean                                       17.81µ        10.89µ       -38.86%

                                           │    old.txt     │                  new.txt                  │
                                           │      B/op      │     B/op      vs base                     │
Checker/name=function_calls-8                  0.000 ± 0%       0.000 ± 0%         ~ (p=1.000 n=10) ¹
Checker/name=unary_and_binary_operations-8     0.000 ± 0%       0.000 ± 0%         ~ (p=1.000 n=10) ¹
Checker/name=deep_struct_access-8            1.242Ki ± 0%     0.000Ki ± 0%  -100.00% (p=0.000 n=10)
geomean                                                   ²                 ?                       ² ³
¹ all samples are equal
² summaries must be >0 to compute geomean
³ ratios must be >0 to compute geomean

                                           │   old.txt    │                 new.txt                 │
                                           │  allocs/op   │ allocs/op   vs base                     │
Checker/name=function_calls-8                0.000 ± 0%     0.000 ± 0%         ~ (p=1.000 n=10) ¹
Checker/name=unary_and_binary_operations-8   0.000 ± 0%     0.000 ± 0%         ~ (p=1.000 n=10) ¹
Checker/name=deep_struct_access-8            159.0 ± 0%       0.0 ± 0%  -100.00% (p=0.000 n=10)
geomean                                                 ²               ?                       ² ³
¹ all samples are equal
² summaries must be >0 to compute geomean
³ ratios must be >0 to compute geomean
```

Nice -38.86% improvments. 👍🏻

diegommm:
> Nice -38.86% improvments. 👍🏻

I updated the benchmarks, now it's ~58%!

> Awesome, it's really interesting to see the results. I think this will be very beneficial for users who use Eval path and do not precompile expressions.

This would also be useful in tight loops when we need to compile many expressions, for example when a service initializes.

checker/nature/nature.go
diegommm: I used your approach in the first version, but it was very difficult to use.
I tried two ways:

- Modify the signature of the methods to pass a `*Cache` as first argument, which can be nil.
- Call `SetCache` every time I needed it to ensure it was used.

The problems I found:

- There is no reason to not use the cache, so I think it would be ok to have it be a part.
- The logic to account for cache not being set every time creates a lot of complexity. Remember this happens for every method. I have to look for a suitable alternative in every case. This makes it hard to follow and verify every case.

The solution I came to:

- The cache is not agent, it is also functional.
- The cache is a proxy but it also makes it easier to centralize all the logic.
- There is a lot of interconnected logic. If this logic is addressed in a centralized way then it is much easier to implement.
- The cache should be local to a `conf.Config`. This is because that's the best way to reuse it throughout the lifetime of a program being compiled. Starting with the type checker, all following compilation steps use it. It is as much important as the environment itself, because it allows us to determine if the environment contains a certain field and how to retrieve it.
- In the early stages when we are parsing the program we are just building the AST so we only care about creating the basics. For example, we don't know yet what type is returned at this point. But then we bind the AST to a cache then it means we are finally interested in type information details.
- Reading myself above, I think "bind" is a better word. I have just changed from `SetCache` to `Bind` to express this relationship.
- The cache is not only meant for the type checker. I only changed mainly the type checker because it's the core, but it is also meant to be used by the optimizer, patchers, and the compiler. I also saw that we could probably use it in the vm and in the runtime because this would avoid some costly reflection checks.
- I didn't go into more usage of it because I made many changes already. I would like to see if these changes are ok and then take a look at the other packages.

> We already have this problem and we have to set a nature cache, for example, in the following line:

So as I explain above, this is not a problem, it's part of the design. In AST we don't care about type information, we only know about raw elements. So when we need more information we need to bind it with `Bind` to a cache.

conf/config.go
diegommm: See the other comment.

checker/nature/nature.go
```diff
@@ -1,261 +1,615 @@
...
+	*Optional
+	*FuncData
```
antonmedv, line 42: Please also write the comment on Why? do we need separate *Optional and *FuncData so reader will understand it.

checker/nature/nature.go
```diff
@@ -1,261 +1,615 @@
...
+		fd = new(FuncData)
+	}
+	return Nature{Type: t, Kind: k, FuncData: fd, cache: c}
 }
 
-func (n Nature) Deref() Nature {
-	if n.Type != nil {
-		n.Type = deref.Type(n.Type)
+func (c *Cache) getStruct(t reflect.Type) Nature {
+	if c != nil {
```
antonmedv, line 120: How can `c` be nil here? Looks like redundent if.

checker/nature/nature.go
```diff
@@ -1,261 +1,615 @@
...
-	case n.Type == nil && !n.Nil:
-		return true
-	case n.IsAny():
-		return true
+// Cache is a shared cache of type information. It is only used in the stages
+// where type information becomes relevant, so packages like ast, parser, types,
+// and lexer do not need to use the cache because they don't need any service
+// from the Nature type, they only describe. However, when receiving a Nature
+// from one of those packages, the cache must be set immediately.
+type Cache struct {
```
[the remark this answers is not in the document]
diegommm: Yes, sounds like a good idea. I will make the cache global.

checker/nature/nature.go
diegommm: There was a case but I moved the check to the top to make this more clear. I also added a comment for it.

diegommm:
Completed refactor responding to all questions.
I updated the benchmarks too, they are similar after some other improvements.

antonmedv:
Nice! Let me take a look at the improved PR.

antonmedv:
Very nice job! 👍🏻

I like the refactored version (with passing cache from the top much better). I think it makes code more readable.

Issue #234, opened 2022-04-27
Enhance expr.Env() options on structs


andrewbenton:
When passing structs to the `expr.Env()` option, it would be nice to have a way to modify the field extraction behavior to be able to:
- extract or map names based on tags such as the common `json` tag
- ignore non-exported fields
- ignore methods

In my particular case, I'm working with protobuf-generated structs and using the `expr` library as a way to implement a filtering grammar similar to what is described in https://google.aip.dev/160.

Here's an example struct:

```go
type MyType struct {
    state protoimpl.MessageState
    sizeCache protoimpl.SizeCache
    unknownFields protoimpl.UnknownFields

    Id         string                 `protobuf:"bytes,1,opt,name=id,proto3" json:"id,omitempty"`
    Name       string                 `protobuf:"bytes,2,opt,name=name,proto3" json:"name,omitempty"`
    CreateTime *timestamppb.Timestamp `protobuf:"bytes,3,opt,name=create_time,json=createTime,proto3" json:"
}

func (x *MyType) GetId() string {
    if x != nil {
        return x.Id
    }
    return ""
}

/* other generated functions for exported fields */
```

The above struct results in both the non-exported fields and duplicate methods being made available to the expression context.  It also means that the default names in the expression (prior to patching) don't match the names given in the protobuf.

antonmedv:
Make sense. I'll add it to my todo list :)

antonmedv:
Done!

NickYadance:
> Done!

Similar issue but failed to find workaround. So is there any way to refer field by json tag in expr for this protobuf case ?

PR #829, opened 2025-09-09
Add support for custom struct tags via environment variable

NickYadance:
Fix: https://github.com/expr-lang/expr/issues/234

Add support for custom struct tags via environment variable, set gotags to read by **system env variables.**

To easily reference field by custom go tags like `json`, `mapstructre`, which is especially useful when dealing with generated go structs like protobuf.

### DIFF
--- a/environment/environment.go
+++ b/environment/environment.go
@@ -1,0 +1,16 @@
+package environment
+
+import "os"
+
+const (
+	ExprEnvGotag = "expr__gotag"
+	DefaultGotag = "expr"
+)
+
+func GetGoTag() string {
+	if v := os.Getenv(ExprEnvGotag); v != "" {
+		return v
+	} else {
+		return DefaultGotag
+	}
+}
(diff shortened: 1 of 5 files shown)


NickYadance:
@antonmedv how do you think about setting custom gotag from system env ?

environment/environment.go
```diff
@@ -0,0 +1,16 @@
...
+
+import "os"
+
+const (
+	ExprEnvGotag = "expr__gotag"
+	DefaultGotag = "expr"
+)
+
+func GetGoTag() string {
+	if v := os.Getenv(ExprEnvGotag); v != "" {
```
antonmedv, line 11: Why do we need to specify this from env?)

vm/runtime/runtime.go
```diff
@@ -65,7 +66,7 @@ func Fetch(from, i any) any {
 		fieldName := i.(string)
 		value := v.FieldByNameFunc(func(name string) bool {
 			field, _ := v.Type().FieldByName(name)
-			switch field.Tag.Get("expr") {
+			switch field.Tag.Get(environment.GetGoTag()) {
```
antonmedv, line 69: Better not introduce another configuration surface and use same conf pgk.

PR #842, opened 2025-09-18
make parser throw error if predicate not present

davixcky:


antonmedv:
I think if function all was disabled, it should be able to be overwritten.

antonmedv:
@davixcky  I will merge this. But will be nice to add tests. Could you please add tests in a separate PR? Thanks

Issue #840, opened 2025-09-18
Field functions after embedded structs are not visible


antonmedv:
@diegommm looks like something to broken with the way nature iterates embedded structs and fields as methods.

```go
type Env struct {
	EmbeddedEnv
	Func func() int
}

type EmbeddedEnv struct{}
```

In this case Func will be not accessible.

antonmedv:
I added the falling test. Please, take a look.

https://github.com/expr-lang/expr/blob/5f4de2012b5ae915bc90db0067201b5f90f24e3f/test/issues/840/issue_test.go#L10-L30

diegommm:
Will take a look.

diegommm:
Created a fix in https://github.com/expr-lang/expr/pull/843

PR #843, opened 2025-09-21
Issue 840: Fix get field pointer struct

diegommm:
Fixes https://github.com/expr-lang/expr/issues/840


diegommm:
@antonmedv I was reviewing the code and remembered this piece that I forgot to mention to you. I created this issue to address it separately: https://github.com/expr-lang/expr/issues/844

PR #845, opened 2025-09-21
Fix #844 alternative 1: remove incorrect bug notice

### LINKED ISSUE #844: Checker not failing when accessing unexported struct fields
When I was working in https://github.com/expr-lang/expr/pull/824, I found what I think is an issue. But there is [this test](https://github.com/expr-lang/expr/blob/69b27afae29804c646a7d51f8daaa032becd799d/checker/checker_test.go#L757-L771) which was failing if I tried to "fix" it. So I thought it was not an issue so I didn't apply a fix, but I would like to discuss it.

I will create to PRs, I will ask you to decide which one we should apply and we can close the other one:

- https://github.com/expr-lang/expr/pull/845 if you want to conserve the current behaviour
- https://github.com/expr-lang/expr/pull/846 if you agree with me

diegommm:
Fix https://github.com/expr-lang/expr/issues/844

### DIFF
--- a/checker/nature/utils.go
+++ b/checker/nature/utils.go
@@ -52,8 +52,6 @@ func (s *structData) structField(c *Cache, parentEmbed *structData, name string)
 	// Lookup own fields first.
 	for ; s.ownIdx < s.numField; s.ownIdx++ {
 		field := s.rType.Field(s.ownIdx)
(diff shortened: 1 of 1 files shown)


antonmedv:
Closing in a favor of https://github.com/expr-lang/expr/pull/846

Issue #831, opened 2025-09-12
Option to disable short-circuiting


lvliangxiong:
```go
package main

import (
	"fmt"

	"github.com/antonmedv/expr"
)

func main() {
	env := map[string]any{
		"a": func() bool { fmt.Println("a executed"); return true },
		"b": func() bool { fmt.Println("b executed"); return false },
	}

	program, _ := expr.Compile(`a() || b()`, expr.Env(env))

	_, _ = expr.Run(program, env)
	// Output:
	// a executed
}

```

Output of above code will be `a executed`. 
But I'm wondering if there's an option which I can pass to the expr, or something else I'can do to disable the shortcircuit effect. 
After disabling shortcircuit, output will be `a executed` and `b executed`.

antonmedv:
Lets add an option to disable short-circuiting `expr.NoShortCircuiting()`. 

In order for this to work compiler should procude different order of opcodes. 

For now, as a workaround, you can use something like this:

```
let r1 = a();
let r2 = b();
r1 || r2
```

lvliangxiong:
I just find a trick which can achieve the same effect:
```go
func ExampleDisableShortCircuit() {
	OR := func(a, b bool) bool {
		return a || b
	}

	env := map[string]any{
		"foo": func() bool {
			fmt.Println("foo")
			return false
		},
		"bar": func() bool {
			fmt.Println("bar")
			return false
		},
		"OR": OR,
	}

	program, _ := expr.Compile("true || foo() or bar()", expr.Env(env), expr.Operator("or", "OR"), expr.Operator("||", "OR"))
	got, _ := expr.Run(program, env)
	fmt.Println(got)

	// Output:
	// foo
	// bar
	// true
}
```

For sure, it would be better we have an Option to do this.

antonmedv:
Will be released with v1.17.7

PR #847, opened 2025-09-25
feat: add disable short-circuiting option to compiler

lvliangxiong:
#831

### DIFF
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -17,6 +17,8 @@ import (
 	"github.com/expr-lang/expr/vm/runtime"
 )
 
+const maxFnArgsBuf = 256
+
 func Run(program *Program, env any) (any, error) {
 	if program == nil {
 		return nil, fmt.Errorf("program is nil")
@@ -82,6 +84,8 @@ func (vm *VM) Run(program *Program, env any) (_ any, err error) {
 	}
 	vm.memory = 0
 	vm.ip = 0
+
+	var fnArgsBuf []any
 
 	for vm.ip < len(program.Bytecode) {
 		if debug && vm.debug {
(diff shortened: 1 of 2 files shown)


lvliangxiong:
I just fix the code according to the check result. It'll be good now.

antonmedv:
Nice! Let me review this more carefully.

diegommm:
We could implement this with less core changes if we follow your example [here](https://github.com/expr-lang/expr/issues/831#issuecomment-3332424208).

My suggestion is the following ([Go Playground](https://go.dev/play/p/PC3JIWUD89E) working example):
```go
// DisableShortCircuit turns short circuiting behaviour off in `&&`, `and`, `||`, and `or` boolean operators.
func DisableShortCircuit() expr.Option {
	return func(c *conf.Config) {
		expr.Function("OR", func(params ...any) (any, error) {
			return params[0].(bool) || params[1].(bool), nil
		}, func(a, b bool) bool { return true })(c)
		expr.Function("AND", func(params ...any) (any, error) {
			return params[0].(bool) && params[1].(bool), nil
		}, func(a, b bool) bool { return true })(c)
		expr.Operator("or", "OR")(c)
		expr.Operator("||", "OR")(c)
		expr.Operator("and", "AND")(c)
		expr.Operator("&&", "AND")(c)
	}
}
```

This way we will not need any changes to vm and compiler packages and the core language remains smaller.

lvliangxiong:
> We could implement this with less core changes if we follow your example [here](https://github.com/expr-lang/expr/issues/831#issuecomment-3332424208).
> 
> My suggestion is the following ([Go Playground](https://go.dev/play/p/PC3JIWUD89E) working example):
> 
> ```go
> // DisableShortCircuit turns short circuiting behaviour off in `&&`, `and`, `||`, and `or` boolean operators.
> func DisableShortCircuit() expr.Option {
> 	return func(c *conf.Config) {
> 		expr.Function("OR", func(params ...any) (any, error) {
> 			return params[0].(bool) || params[1].(bool), nil
> 		}, func(a, b bool) bool { return true })(c)
> 		expr.Function("AND", func(params ...any) (any, error) {
> 			return params[0].(bool) && params[1].(bool), nil
> 		}, func(a, b bool) bool { return true })(c)
> 		expr.Operator("or", "OR")(c)
> 		expr.Operator("||", "OR")(c)
> 		expr.Operator("and", "AND")(c)
> 		expr.Operator("&&", "AND")(c)
> 	}
> }
> ```
> 
> This way we will not need any changes to vm and compiler packages and the core language remains smaller.

I also considered this solution at first glance. However, after spending some time investigating the implementation, I realized this more intuitive solution might potentially conflict with or be overridden by other patch/optimization features. I'm not sure if my concerns are unnecessary, so I've implemented this in the compiler & vm package, which I find easier to understand and might be more robust without bugs.

On the other hand, I give it a try and find something interesting. `Function` here needs type information to be matched with operator. So if we don't give the Env obj when compiling, this option might not take effect, illustrated by the example below:

```
func TestDisableShortCircuit(t *testing.T) {
	count := 0
	exprStr := "foo() or bar()"
	env := map[string]any{
		"foo": func() bool {
			count++
			return true
		},
		"bar": func() bool {
			count++
			return true
		},
	}

	// You must set expr.Env(env) to make expr.DisableShortCircuit() to be effective.
	program, _ := expr.Compile(exprStr, expr.DisableShortCircuit(), expr.Env(env))
	got, _ := expr.Run(program, env)
	assert.Equal(t, 2, count)
	assert.True(t, got.(bool))

	program, _ = expr.Compile(exprStr)
	got, _ = expr.Run(program, env)
	assert.Equal(t, 3, count)
	assert.True(t, got.(bool))
}
```
Since I don't have much time to figure out every detail of the implementation, I'm not sure if this operator overloading solution has other hidden bugs or limitations, or we can solve this by some other tricks? What do you think about this? Looking forward to your reply.

antonmedv:
I think proposal solution in this poor request is proper. Disable short circuit should generate different bytecode.

lvliangxiong:
Any progress here? Let me know if I can help.

antonmedv (approved):

PR #864, opened 2025-11-27
fix(optimizer): wrap merged results as predicates

### LINKED ISSUE #857: Interface conversion panic
Given
```
let foo = {
  "entry": {
    "alpha": "x",
    "beta": 1
  },
};

let bar = {
  "entry": {
    "alpha": "x",
    "beta": 1
  },
};
```
The following generates a panic with the message `interface conversion: ast.Node is *ast.BinaryNode, not *ast.PredicateNode`
```
let bad = foo
  | keys()
  | filter(# in bar)
  | filter(foo[#].alpha == bar[#].alpha)
  | filter(foo[#].beta == bar[#].beta);
bad
```
Whereas this works as expected
```
let bad = foo
  | keys()
  | filter(# in bar)
  | filter(foo[#].alpha == bar[#].alpha);

bad | filter(foo[#].beta == bar[#].beta)
```

thevilledev:
Ensure combined predicates in consecutive `filter()` calls are wrapped in an `ast.PredicateNode`. Previously, the optimizer created a raw `ast.BinaryNode` for merged filters, causing type assertion panics. The expr parser generates `filter(array, predicate)` calls only.

This fix ensures that the merged predicate is wrapped as an `ast.PredicateNode`. Adds a regression test.

Fixes #857

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -3,6 +3,7 @@ package builtin
 import (
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"reflect"
 	"sort"
@@ -16,6 +17,10 @@ import (
 var (
 	Index map[string]int
 	Names []string
+
+	// MaxDepth limits the recursion depth for nested structures.
+	MaxDepth      = 10000
+	ErrorMaxDepth = errors.New("recursion depth exceeded")
 )
 
 func init() {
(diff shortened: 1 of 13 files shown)


thevilledev:
Am I reading this right that this also addresses the OSS-Fuzz finding in https://github.com/expr-lang/expr/issues/790?

btw @antonmedv if you're looking for more eyes on the OSS-Fuzz findings in https://github.com/google/oss-fuzz/blob/master/projects/expr/project.yaml I'd be happy to join 👍

antonmedv:
> btw @antonmedv if you're looking for more eyes on the OSS-Fuzz findings in https://github.com/google/oss-fuzz/blob/master/projects/expr/project.yaml I'd be happy to join 👍

Will be nice! Thanks. I will discuss it with other Expr contributors and with Google.

PR #865, opened 2025-11-27
fix(checker): array type in conditional op

### LINKED ISSUE #819: Bug when running expressions where a variable can be either an array or array of arrays.
The following expression should print 1. instead it returns an error
```
let a = [1];
let b = type(a[0]) == 'array' ? a : [a];
b[0][0]
```

Another example:
```
let range = [1,1000];
let arr = false ? range : [range];
map(arr, {len(#)})
```

On the other hand, this works:
```
let range = [1,1000];
let arr = false ? 5 : [range];
map(arr, {len(#)})
```

thevilledev:
Correctly handle mixed array types (e.g. `[]int` vs `[][]int`) in ternary expressions by falling back to generic array type when element types differ.

Fixes #819

### DIFF
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -1277,6 +1277,13 @@ func (v *Checker) conditionalNode(node *ast.ConditionalNode) Nature {
 		return v.config.NtCache.NatureOf(nil)
 	}
 	if t1.AssignableTo(t2) {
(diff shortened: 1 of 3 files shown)


antonmedv:
Nice!

PR #872, opened 2025-12-05
fix: panic when using pointer to map as env

### LINKED ISSUE #825: Panic when using pointer-type map as environment
### 📝Bug Description
The conf.Env() function panics when passed a pointer to a map (e.g., *map[string]int ) instead of a direct map value.
However, from the code perspective, the pointer type is intended to be supported.
```go
// file: conf/env.go
// line: 25
...
v := reflect.ValueOf(env)
d := deref.Value(v)
...

// file: internal/deref/deref.go
// line: 39
func Value(v reflect.Value) reflect.Value {
	for v.Kind() == reflect.Ptr || v.Kind() == reflect.Interface {
		if v.IsNil() {
			return v
		}
		v = v.Elem()
	}
	return v
}
```

### 🌍 Environment
- expr version : Latest (master branch)
- Go version : 1.24
- OS : All platforms

### 🔄 Steps to Reproduce
```go
package main

import (
    "github.com/expr-lang/expr"
)

func main() {
    m := map[string]int{"foo": 42}

    env := &m

    _, err := expr.Compile("foo > 0", expr.Env(env))
    
    if err != nil {
        panic(err)
    }
}
```

### ✅ Expected Behavior
The function should handle pointer-to-map types gracefully by properly dereferencing the pointer before performing map operations. 🎯

### ❌ Actual Behavior
```text
panic: reflect: call of reflect.Value.Len on ptr to non-array Value

goroutine 1 [running]:
reflect.Value.lenNonSlice({0x1011f8ca0?, 0x140000a4378?, 0x1400009c998?})
        /usr/local/go/src/reflect/value.go:1780 +0x220
reflect.Value.Len(...)
        /usr/local/go/src/reflect/value.go:1761
github.com/expr-lang/expr/conf.Env({0x1011f8ca0, 0x140000a4378})
        /Users/bytedance/go/pkg/mod/github.com/expr-lang/expr@v1.17.6/conf/env.go:38 +0x208
github.com/expr-lang/expr/conf.(*Config).WithEnv(0x140000de000, {0x1011f8ca0?, 0x140000a4378?})
        /Users/bytedance/go/pkg/mod/github.com/expr-lang/expr@v1.17.6/conf/config.go:64 +0x58
main.main.Env.func1(0x101219660?)
        /Users/bytedance/go/pkg/mod/github.com/expr-lang/expr@v1.17.6/expr.go:31 +0x24
github.com/expr-lang/expr.Compile({0x101187670, 0x7}, {0x1400009cf08, 0x1, 0x1400005c738?})
        /Users/bytedance/go/pkg/mod/github.com/expr-lang/expr@v1.17.6/expr.go:211 +0x17c
main.main()
        /Users/bytedance/projects/go/rule/main.go:12 +0xc8
exit status 2

```

### 🔍 Root Cause Analysis
In conf/env.go , lines 38-43, the code uses:

- v.Len() ❌ fails on pointer types
- v.MapKeys() ❌ fails on pointer types
- v.MapIndex(key) ❌ fails on pointer types

The fix is to use the dereferenced value d (from deref.Value(v) ) for these operations. 💡

I have a working [implementation](https://github.com/yawnBright/expr/commit/a8b13f67affe44ab16a3ed16e31e40092ecc7703) ready that addresses this issue. Would you like me to submit a PR for this fix?

thevilledev:
Previously, passing a pointer to a map (e.g. `*map[string]any`) to `expr.Env()` would cause a panic because reflection methods were called on the pointer value instead of the underlying map. With this change the value is first deref'd in the config before any map operations are performed.

Consequently, the `vm.Run()` method needs to dereference the env map. This ensures compatibility with the optimisations like `OpLoadFast` where the compiler determines the environment simply as `map[string]any`.

Regression test added.

Fixes #825.

### DIFF
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -462,7 +462,7 @@ func (v *Checker) binaryNode(node *ast.BinaryNode) Nature {
 				return v.error(node, err.Error())
 			}
 		}
-		if l.IsString() && r.IsString() {
+		if (l.IsString() || l.IsByteSlice()) && r.IsString() {
 			return v.config.NtCache.FromType(boolType)
 		}
 		if l.MaybeCompatible(&v.config.NtCache, r, StringCheck) {
@@ -549,6 +549,13 @@ func (v *Checker) memberNode(node *ast.MemberNode) Nature {
 
 	switch base.Kind {
 	case reflect.Map:
+		// If the map key is a pointer, we should not dereference the property.
+		if !prop.AssignableTo(base.Key(&v.config.NtCache)) {
+			propDeref := prop.Deref(&v.config.NtCache)
+			if propDeref.AssignableTo(base.Key(&v.config.NtCache)) {
+				prop = propDeref
+			}
+		}
 		if !prop.AssignableTo(base.Key(&v.config.NtCache)) && !prop.IsUnknown(&v.config.NtCache) {
 			return v.error(node.Property, "cannot use %s to get an element from %s", prop.String(), base.String())
 		}
@@ -562,6 +569,7 @@ func (v *Checker) memberNode(node *ast.MemberNode) Nature {
 		return base.Elem(&v.config.NtCache)
 
 	case reflect.Array, reflect.Slice:
+		prop = prop.Deref(&v.config.NtCache)
 		if !prop.IsInteger && !prop.IsUnknown(&v.config.NtCache) {
 			return v.error(node.Property, "array elements can only be selected using an integer (got %s)", prop.String())
 		}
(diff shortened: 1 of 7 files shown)


antonmedv:
But why would `*map[string]any` pointer to map is needed?

thevilledev:
> But why would `*map[string]any` pointer to map is needed?

I guess this would mostly be a quality of life improvement. Tried to think of an example. Maybe you unmarshal the `env` from JSON and compile a rule, and you use a reference to the map by thinking "well why not, I have it right here". One possible example (which panics): https://go.dev/play/p/WsuJJtq-DwL

Also @yawnBright feel free to chime in with your use case.

yawnbright:
> "well why not, I have it right here".

I think that's how I found this issue

antonmedv:
Let's say for now we do not support this feature. It should be struct, pointer to the struct, or a map.

Issue #777, opened 2025-03-30
[Request/Question] Support proxies for `expr.Env`


x1unix:
Greetings.

I've a question - does `expr` support having proxies as `expr.Env` (something similar to [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) feature in JS)?

I have a situation where I've to deal with a chain of objects as env that have prototype inheritance (also like in JS).

As env itself is dynamic (might depend on context, environment vars, etc), in order to build `expr.Env` each time I need to run an expression - it's necessary to traverse a whole prototype chain and create a new map.

### Proposed solution

Ideally, to handle such cases, `expr.Env` could consume a `ProxyEnv` interface that will be called to get a variable:

```go
// feel free to provide a better name
type ProxyEnv interface {
   GetValue(path []string) (bool, any)
}
```

The interface could receive a path so it can support deep queries (e.g. `foo.bar.baz`).

### Pros

This feature will allow to handle complex cases such as:

* Cases when value have to be pulled dynamically in runtime.
* To abstract away env when it has a complex structure, such as prototype chain.
* Avoid allocating additional maps or structs for env.

antonmedv:
I think we can add those features. 


What about this case:

     foo.bar[42 + x].baz

x1unix:
@antonmedv good question!

Please check updated proposal.

x1unix:
@antonmedv if you're interested, I can try to make a PR after the proposal will be finalized.

antonmedv:
I think implementation should be done via patcher which will replace property access with function cause.

x1unix:
@antonmedv if patched code will throw an error, would error position point to an original (pre-patched) code?

Let's imagine a code like this:

```js
foo.bar.baz
```

and patched version would look like this:

```js
foo.bar().baz() // bar returns nil
```

Ideally returned `EvalError` should be transparently mapped to original code:

```
foo.bar.baz
            ^^^ Cannot read properties of undefined (reading "baz")
```

antonmedv:
Expr patcher designed to point to original code location! Actually a lot of optimizations is done via patcher. But errors always return original code allocation.

PR #873, opened 2025-12-06
feat: implement proxies support

x1unix:
This PR adds support of proxies. Proxies allow intercepting property access, like Python's `__get__` magic method.

Proxies allow implementing different patterns like prototype chaining, read on demand, etc.

Fixes: https://github.com/expr-lang/expr/issues/777

### DIFF
--- a/vm/runtime/runtime.go
+++ b/vm/runtime/runtime.go
@@ -13,17 +13,12 @@ import (
 // Proxy is an interface that allows intercepting object property access.
 type Proxy interface {
 	// GetProperty returns the value of the property with the given key.
-	GetProperty(key any) (any, bool)
+	GetProperty(key any) any
 }
 
 func Fetch(from, i any) any {
 	if proxy, ok := from.(Proxy); ok {
-		r, ok := proxy.GetProperty(i)
-		if !ok {
-			panic(fmt.Sprintf("cannot fetch %v from proxy", i))
-		}
-
-		return r
+		return proxy.GetProperty(i)
 	}
 
 	v := reflect.ValueOf(from)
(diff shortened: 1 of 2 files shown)


vm/runtime/runtime.go
```diff
@@ -10,7 +10,17 @@ import (
...
 )
 
+// Proxy is an interface that allows intercepting object property access.
+type Proxy interface {
+	// GetProperty returns the value of the property with the given key.
+	GetProperty(key any) any
+}
+
 func Fetch(from, i any) any {
+	if proxy, ok := from.(Proxy); ok {
```
antonmedv: Relying only on real time to check for proxy interface is not the best strategy. We need to do a pre-optimization and check at compile time whether a special opcode for fetching using this interface can be inserted ahead of time. With just a runtime check for proxy, I cannot accept this pull request.

x1unix: I've a couple of questions:

- By ahead of time optimization, do you mean to implement env check in `optimizer.Optimize` method?
  - Current API allows using different envs for parsing, optimization and evaluation. 
  This means that during eval stage, user can supply an env that may have a proxy (e.g. `foo.bar.baz`), but this wasn't accounted during parse/optimize stage, so iface check still might be required inside `OpFetch`.
  - Proxy may be located deeply inside env, or returned from a func.
  - Initially, I wanted to use type check instead of an interface as it is 2x cheaper, but couldn't find a suitable approach (see [benchmark](https://gist.github.com/x1unix/b710aa92735aed446ee76c7e0d6c56ca)). Happy to accept different proposals.
  - As 90% of users probably won't need proxies, this feature can be opt-in (feature flag). This will keep perf the same as before and will skip proxy checks by default.

antonmedv:
As the current implementation greatly affects Fetch runtime, i'm closing this PR. Feel free to provide better implementation for those "proxies".

Issue #854, opened 2025-10-20
Bug in optional chaining typechecker


antonmedv:
Example from playground:

```
user.Profile?.Address ?? "Unknown address"
```

```
user:
  Id: 1
  Name: John Doe
  Group: admin
```

ehabterra:
Hey @antonmedv, 

I'd like to contribute to this project and this issue looks like a good fit. I'm still learning the codebase, so please bear with me.

I've been investigating the issue. For the expression `user.Profile?.Address ?? "Unknown address"`, when Profile doesn't exist on the User struct, the type checker throws a `"has no field"` error during type checking.

Is this the issue we need to fix?

I also have a question about scope: should we also handle cases like:

```go
user.Profiles[0]?.Address ?? "Unknown address"
```

I've made some changes to the checker, but I'm encountering an error from `runtime.go` file: `"cannot fetch Profile from expr_test.User (1:6)"`. I'm still debugging this and would appreciate any guidance on the expected behavior.

Thanks!

antonmedv:
Hey @ehabterra 

Yes, i think this is the issue. Now sure what is the culprit here. Need to investigate.

ehabterra:
This should be resolved now. Please let me know if I missed anything or if there's anything I should address. #874

antonmedv:
Actually, behavior here is correct. Playground is using types.Map and this is proper behavior.

PR #874, opened 2025-12-08
fix: bug in optional chaining typechecker #854

ehabterra:
#854 fix bug in optional chaining typechecker:
- Add property for `Optional` property in `StringNode`.
- Add property to `Skip` compiling in `Nature`.
- Add `TODO` for `user.Profile[0]?.Address ?? "Unknown address"` example.


antonmedv:
Fixed in https://github.com/expr-lang/expr/commit/f493d7dbaef0b18fe2cf578b1771a2f68dd03c0b

PR #883, opened 2025-12-16
fix(patcher): ctx into nested custom funcs and Env

### LINKED ISSUE #823: WithContext doesn't work with 2 functions within 1 call
Hi @antonmedv 

If there are 2 patched function calls within 1 function call, it doesn't compile.

https://go.dev/play/p/9dH2e9KuF2d

```go
package main

import (
	"context"
	"fmt"
	"time"

	"github.com/expr-lang/expr"
)

type env struct {
	Ctx context.Context `expr:"ctx"`
}

func main() {
	p, err := expr.Compile(
		"now2().After(date2())",
		expr.Env(env{}),
		expr.WithContext("ctx"),
		expr.Function(
			"now2",
			func(params ...any) (any, error) { return time.Now(), nil },
			new(func(context.Context) time.Time),
		),
		expr.Function(
			"date2",
			func(params ...any) (any, error) { return time.Date(2000, 1, 1, 0, 0, 0, 0, time.UTC), nil },
			new(func(context.Context) time.Time),
		),
	)
	if err != nil {
		panic(err)
	}
	r, err := expr.Run(p, &env{Ctx: context.Background()})
	if err != nil {
		panic(err)
	}
	fmt.Println(r)
}
```

Output:
```
panic: not enough arguments to call date2 (1:14)
	 | now2().After(date2())
	 | .............^

goroutine 1 [running]:
main.main()
	/tmp/sandbox2261374222/prog.go:32 +0x2d5
```

thevilledev:
WithContext patcher now looks up function types from the Functions table and Env methods when the callee type is `interface{}`. This fixes context injection for custom functions and Env methods nested as arguments inside method calls with unknown callee types (e.g., `Now2().After(Date2()))`.

Also improved the regression test to actually verify context is passed to both functions, and added a test for Env methods.

Should finally fix #823.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-versions: [ '1.18', '1.22', '1.24', '1.25' ]
+        go-versions: [ '1.18', '1.22', '1.24', '1.25', '1.26' ]
         go-arch: [ '386' ]
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-versions: [ '1.18', '1.19', '1.20', '1.21', '1.22', '1.23', '1.24', '1.25' ]
+        go-versions: [ '1.18', '1.19', '1.20', '1.21', '1.22', '1.23', '1.24', '1.25', '1.26' ]
     steps:
       - uses: actions/checkout@v3
       - name: Setup Go ${{ matrix.go-version }}
--- a/README.md
+++ b/README.md
@@ -1,8 +1,5 @@
 <h1><a href="https://expr-lang.org"><img src="https://expr-lang.org/img/logo.png" alt="Zx logo" height="48"align="right"></a> Expr</h1>
 
-> [!IMPORTANT]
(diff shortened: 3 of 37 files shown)


thevilledev:
I had an idea to maybe fix this through a checker-level fix but it turned out to be quite cumbersome due to side-effects to all the other expressions. Lemme know what you think @antonmedv :+1:

antonmedv (approved):

patcher/with_context.go
```diff
@@ -19,6 +21,16 @@ func (w WithContext) Visit(node *ast.Node) {
 		if fn == nil {
 			return
 		}
+		// If callee type is interface{} (unknown), try to look up the function type
+		// from the Functions table. This handles cases where the call is nested
+		// inside another call with an unknown callee type.
+		if fn.Kind() == reflect.Interface && w.Functions != nil {
```
antonmedv: But why date2() callnode type is any?

thevilledev: IIUC the checker returns early without visiting `date2()` because `now2()` initially has 0 arguments, as the context has not been injected yet. This causes `checkArguments` to fail. Furthermore it makes the `now2()` call return unknown type. Then it cascades: memberNode for `.After` returns unknown, so the outer callNode returns early without visiting its arguments. And it leaves `date2()` identifier with `interface{}` type when WithContext runs.

I was playing around with the Checker-level fix like:

```go
if nt.IsUnknown(&v.config.NtCache) {
    // Still visit arguments to set their types for patchers
    for _, arg := range node.Arguments {
        v.visit(arg)
    }
    return Nature{}
}
```

But it breaks a bunch of stuff, and I went with this. :)

I'll update the comment a bit.

antonmedv: Thanks for explanation, let me also see if checker solution can be added.

PR #908, opened 2026-01-19
perf(vm): optimize loop iteration with scope pool

thevilledev:
## Motivation

Iteration over slices previously used reflection to access elements, which was slow and allocated unnecessarily. 

## Changes

This change adds type-specialized fast paths for common slice types (`[]int`, `[]float64`, `[]string`, `[]any`) that bypass reflection entirely. Scope objects are now pooled and reused across loop iterations. The current scope pointer is cached to avoid repeated slice lookups.

Bench run:

```bash
go test -bench='Benchmark_filter' -benchmem -count=10 -run=^$ ./...
```

Results:

```
cpu: Apple M1
               │   old.txt   │               new.txt               │
               │   sec/op    │   sec/op     vs base                │
_filter-8        46.70µ ± 1%   39.35µ ± 0%  -15.73% (p=0.000 n=10)
_filterLen-8     42.98µ ± 0%   36.62µ ± 0%  -14.79% (p=0.000 n=10)
_filterFirst-8   408.4n ± 0%   337.2n ± 0%  -17.42% (p=0.000 n=10)
_filterLast-8    700.7n ± 0%   681.2n ± 1%   -2.78% (p=0.000 n=10)
_filterMap-8     4.970µ ± 0%   3.611µ ± 0%  -27.35% (p=0.000 n=10)
geomean          4.910µ        4.126µ       -15.98%

               │   old.txt    │               new.txt                │
_filter-8        20.38Ki ± 0%   18.20Ki ± 0%  -10.69% (p=0.000 n=10)
_filterLen-8     7.938Ki ± 0%   6.039Ki ± 0%  -23.92% (p=0.000 n=10)
_filterFirst-8     192.0 ± 0%     224.0 ± 0%  +16.67% (p=0.000 n=10)
_filterLast-8      712.0 ± 0%     808.0 ± 0%  +13.48% (p=0.000 n=10)
_filterMap-8      1736.0 ± 0%     920.0 ± 0%  -47.00% (p=0.000 n=10)
geomean          2.045Ki        1.763Ki       -13.77%

               │   old.txt    │               new.txt                │
               │  allocs/op   │ allocs/op   vs base                  │
_filter-8         1155.0 ± 0%   864.0 ± 0%  -25.19% (p=0.000 n=10)
_filterLen-8      1004.0 ± 0%   749.0 ± 0%  -25.40% (p=0.000 n=10)
_filterFirst-8    12.000 ± 0%   4.000 ± 0%  -66.67% (p=0.000 n=10)
_filterLast-8      24.00 ± 0%   24.00 ± 0%        ~ (p=1.000 n=10) ¹
_filterMap-8     123.000 ± 0%   9.000 ± 0%  -92.68% (p=0.000 n=10)
geomean            132.7        56.17       -57.66%
¹ all samples are equal
```

## Further comments

As the `Scope` struct is bigger, so is the B/op for filterFirst and filterLast. They are faster still.


thevilledev:
The fuzzing finding here looks to be valid but unrelated to this change. I'll spawn another PR from this.

antonmedv:
Nice! Another thing what will be cool to investigate is how to iterate []struct arrays. This is the most common operation. Any possibility is using pointer arithmetic (maybe behind a build flag)

thevilledev:
Interesting idea, so basically store `unsafe.Pointer` in `Scope` struct? Will be interesting to see how that could play out

antonmedv:
Additionally would be really nice to think how we can improve the cases for `[]any`.

PR #909, opened 2026-01-21
perf(builtin): min, max, mean, median fast paths

thevilledev:
## Motivation

The `min`, `max`, `mean`, and `median` builtin functions were using reflection for every element in the input array, calling `reflect.Value.Index(i).Interface()` in a loop. This causes significant allocations (100+ per call for 100 elements) because each `.Interface()` call triggers heap allocations.

## Changes
Avoid reflection and per-element allocations for common typed slices (`[]int`, `[]float64`, `[]any`) by adding type-switch fast paths that iterate directly without calling `reflect.Value.Interface()`.

Added benchmarks:

- `Benchmark_min`
- `Benchmark_max`
- `Benchmark_mean`
- `Benchmark_median`

Benchstat results against `master` branch:

```
cpu: Apple M1
          │  master.out  │               fix.out               │
          │    sec/op    │   sec/op     vs base                │
_min-8      2850.5n ± 0%   427.7n ± 1%  -85.00% (p=0.000 n=10)
_max-8      2877.5n ± 1%   443.2n ± 0%  -84.60% (p=0.000 n=10)
_mean-8     2140.0n ± 1%   227.2n ± 0%  -89.38% (p=0.000 n=10)
_median-8    4.801µ ± 1%   1.360µ ± 1%  -71.67% (p=0.000 n=10)
geomean      3.030µ        491.9n       -83.76%

          │  master.out  │               fix.out                │
_min-8      1648.00 ± 0%     48.00 ± 0%  -97.09% (p=0.000 n=10)
_max-8      1648.00 ± 0%     48.00 ± 0%  -97.09% (p=0.000 n=10)
_mean-8     1656.00 ± 0%     56.00 ± 0%  -96.62% (p=0.000 n=10)
_median-8   4.391Ki ± 0%   2.047Ki ± 0%  -53.38% (p=0.000 n=10)
geomean     2.071Ki          128.2       -93.95%

          │  master.out  │              fix.out               │
          │  allocs/op   │ allocs/op   vs base                │
_min-8      102.000 ± 0%   2.000 ± 0%  -98.04% (p=0.000 n=10)
_max-8      102.000 ± 0%   2.000 ± 0%  -98.04% (p=0.000 n=10)
_mean-8     103.000 ± 0%   3.000 ± 0%  -97.09% (p=0.000 n=10)
_median-8    211.00 ± 0%   11.00 ± 0%  -94.79% (p=0.000 n=10)
geomean       122.6        3.390       -97.24%
```

## Further comments

Other typed slices (e.g., `[]int32`, `[]uint64`, custom numeric types) fall back to the original reflection-based path. Open to feedback whether we should include more typed slices here.

Similar optimizations could potentially be applied to other builtins like `sum`.

### DIFF
--- a/builtin/builtin_test.go
+++ b/builtin/builtin_test.go
@@ -188,6 +188,8 @@ func TestBuiltin(t *testing.T) {
 		{`reduce(1..9, # + #acc)`, 45},
 		{`reduce([.5, 1.5, 2.5], # + #acc, 0)`, 4.5},
 		{`reduce([], 5, 0)`, 0},
+		{`reduce(10..1, # + #acc, 100)`, 100},
+		{`reduce([], # + #acc, 42)`, 42},
 		{`concat(ArrayOfString, ArrayOfInt)`, []any{"foo", "bar", "baz", 1, 2, 3}},
 		{`concat(PtrArrayWithNil, [nil])`, []any{42, nil}},
 		{`flatten([["a", "b"], [1, 2]])`, []any{"a", "b", 1, 2}},
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -1124,9 +1124,18 @@ func (c *compiler) BuiltinNode(node *ast.BuiltinNode) {
 			c.derefInNeeded(node.Arguments[2])
 			c.emit(OpSetAcc)
 		} else {
+			// When no initial value is provided, we use the first element as the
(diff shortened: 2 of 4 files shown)


antonmedv:
Nice! I like such optimizations!

thevilledev:
Thanks! So much fun :)

Needs more tests and fine-tuning, but I'll mark it ready once there 🤞

Issue #841, opened 2025-09-18
expr.Disabe*() functions are not working


davixcky:
I want to disable some functions, but seems like is not working

```go
	program, err := expr.Compile(query, expr.Option(func(c *conf.Config) {
		c.Disabled["all"] = true

		for name := range c.Disabled {
			fmt.Println("Disabled...", name)
		}
	}))
	if err != nil {
		return nil, fmt.Errorf("error compiling query: %v", err)
	}

```

I even tested with the builtin function, but doesn't work
```go
./cmd  -q "all(sandboxes, .Name startsWith 'frontend')"
Compiling query... 2 all(sandboxes, .Name startsWith 'frontend') for entity type... sandbox
Options... [0x5f3ae0 0x5f3ac0 0x5f3aa0 0x5f4180 0x5f3a40]
Disabled... all
false
```

**Expected**
To see an error `expr.Compile`

antonmedv:
Well, disabling built-in predicates is kind of its own separate feature.

PR #924, opened 2026-02-06
Parser should not throw an error if predicate not present

sc07kvm:
# Context

A change introduced in #842 (related to #841) altered the previous behavior. The earlier behavior was intentional: functions can be provided via env at runtime.

If the following code works:
```go
program, err := expr.Compile(`fn(1)`)
out, err := expr.Run(program, map[string]any{"fn": func(a int) int { return a }})
```
then this code should work as well:
```go
program, err := expr.Compile(`upper(1)`, expr.DisableBuiltin("upper"))
out, err := expr.Run(program, map[string]any{"upper": func(a int) int { return a }})
```

# Compatibility
#842 also introduced a backward-incompatible behavior in a minor release.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-versions: [ '1.18', '1.22', '1.24', '1.25' ]
+        go-versions: [ '1.18', '1.22', '1.24', '1.25', '1.26' ]
         go-arch: [ '386' ]
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


antonmedv:
Alright! Nice lets me review the code.

antonmedv:
I've reverted the change in https://github.com/expr-lang/expr/commit/552eb1baa42c6d65a9de22f65b2922fdc145541f

Also added the regression test in https://github.com/expr-lang/expr/commit/40bda0b3736f9fc5d96c02f970d0ef55748c3ed9 (As I think we do not need then amount of tests, I also see repeating tests cases in current PR).

I will release this changes shortly after.

PR #938, opened 2026-02-26
builtin: enforce MemoryBudget for split and splitAfter

thesmartshadow:
### Summary
Fixes a **MemoryBudget bypass** where allocation-heavy builtins compiled via `OpCall*` were not accounted, allowing large heap allocations despite a small `vm.VM.MemoryBudget`.

### What changed
- Convert `split` and `splitAfter` from `Func` → `Safe` so allocations are charged via `OpCallSafe` + `vm.memGrow`.
- Add regression tests to ensure `MemoryBudget` blocks large `split/splitAfter` outputs.

### Why
`split/splitAfter` can produce very large slices from attacker-controlled input (common in rule engines / policy evaluators). With this change, those allocations are properly budgeted and will return `memory budget exceeded` when the budget is too small.

### Tests
- `go test ./...`
- Added: `TestMemoryBudget_SplitBuiltin`, `TestMemoryBudget_SplitAfterBuiltin`

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -263,14 +263,17 @@ var Builtins = []*Function{
 	},
 	{
 		Name: "split",
-		Func: func(args ...any) (any, error) {
+		Safe: func(args ...any) (any, uint, error) {
+			var parts []string
 			if len(args) == 2 {
-				return strings.Split(args[0].(string), args[1].(string)), nil
+				parts = strings.Split(args[0].(string), args[1].(string))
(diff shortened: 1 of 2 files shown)


sanmaxdev (commented):
The `Safe` path charges each produced element through `memGrow`, and the regression cases cover both split variants under a constrained budget.

Verified:
- `go test ./...`
- `go test -race .`
- `GOARCH=386 go build`

PR #943, opened 2026-03-07
fix(builtin): guard against missing args in builtins

thevilledev:
## Motivation

Relates to #941.

OSS-Fuzz finding has a fuzzed expression which calls `join()` with zero arguments inside a method call on an any-typed value (e.g. `fn().N(join())`). The type checker's `callNode()` skips argument validation when the callee resolves to an unknown type, so the invalid call passes compilation. At runtime `join()` accesses `args[0]` on an empty slice, producing a Go runtime panic instead of a clean error.

I found similar issues from other builtins, relying solely on compile-time validation that can be bypassed.

## Changes

Add runtime `len(args)` guards to a number of builtin functions. Each now returns a descriptive error. This error is already part of the allowlist in the fuzzing harness.

I also removed some autogenerated test lines that called `first` and `last` with wrong arity. These tests were passing because of `defer recover()` silently swallowing the panic.

## Further comments

I think there is a chance to fix this in the checker, to provide compile-time rejection. But I guess it'll be quite complicated because of optional chaining on unknown types for a number of different expressions. This fix is the conservative kind just to have clean and descriptive runtime errors.

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -222,6 +222,9 @@ var Builtins = []*Function{
 	{
 		Name: "trimPrefix",
 		Func: func(args ...any) (any, error) {
+			if len(args) == 0 {
+				return nil, fmt.Errorf("invalid number of arguments (expected at least 1, got 0)")
+			}
 			s := " "
 			if len(args) == 2 {
 				s = args[1].(string)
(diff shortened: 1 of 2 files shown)


builtin/builtin.go
```diff
@@ -312,6 +318,9 @@ var Builtins = []*Function{
 	{
 		Name: "repeat",
 		Safe: func(args ...any) (any, uint, error) {
+			if len(args) < 2 {
```
sanmaxdev, line 321: The checker bypass applies to excess arguments too, but this `< 2` guard accepts them silently. On this head, calling `repeat("x", 2, "extra")` returns `"xx"` with no error; `indexOf`, `join`, and `trimPrefix` behave similarly. Could these guards enforce the exact allowed arity (`!= 2`, or `len < 1 || len > 2` for optional forms) and add an excess-argument regression case?

sanmaxdev (commented):
The missing-argument paths are covered and the full test suite passes. I found one remaining arity gap in the new guards.

PR #945, opened 2026-03-19
fix(parser): support optional chaining on slice expressions

### LINKED ISSUE #822: Unexpected behaviour - cannot slice unkown with ?.
I wouldn't say that this is a bug, but an unexpected behaviour I have come across:

`let x = "test"; x?.[0:1]`

Returns "t", as expected.

`let x = nil; x?.[0:1]`

Returns "error cannot slice unknown (1:17)".  I expected nil.

Possibly the first expression is misuse of the langauge and should not be allowed.  But if it is to be allowed, then I think the consistent result of the second expression should be nil.

eyupcanakman:
`foo?.[0:1]` panics at runtime when `foo` is nil. Optional chaining works for member access (`foo?.bar`) through `MemberNode.Optional` + `OpJumpIfNil`, but `SliceNode` had no equivalent.

Added the same `Optional` + `OpJumpIfNil` pattern to `SliceNode`.

Fixes #822


checker/checker.go
```diff
@@ -613,6 +613,9 @@ func (v *Checker) sliceNode(node *ast.SliceNode) Nature {
 	case reflect.String, reflect.Array, reflect.Slice:
 		// ok
 	default:
+		if node.Optional {
```
sanmaxdev: This suppresses the checker error for every concrete non-slice type, not only nil. Since `OpJumpIfNil` only skips nil values, `let x = 1; x?.[0:1]` compiles and then fails during execution with `cannot slice 0`. Please preserve the compile-time error for concrete non-slice values and add a regression case.

sanmaxdev (changes requested):
The nil case works, but this also moves concrete non-slice values past the checker. I reproduced it with `let x = 1; x?.[0:1]`: compilation succeeds, then execution returns `cannot slice 0`. Please keep concrete non-slice values as compile-time errors and add a regression case.

checker/checker.go
eyupcanakman: The checker now only skips the error for a nil nature, so concrete non-slice types fail at compile time again. Regression cases are in `test/issues/822`.

PR #947, opened 2026-03-20
feat(builtin): add merge() for combining maps

### LINKED ISSUE #895: Add new builtin for merging maps
Expr currently does not have a builtin for merging maps. Adding a new builtin would be useful to merge maps but also to modify nested fields.

For instance:

```javascript
/*
Env:
{
  "data":{
    "status": 0,
    "rows": 32,
    "signing_keys":[
      "key1",
      "key2"
    ]
  }
}
*/

merge($env["data"], {signing_keys: concat($env["data"]["signing_keys"], ["another"])})

/*
Output:
{
  "data":{
    "status": 0,
    "rows": 32,
    "signing_keys":[
      "key1",
      "key2",
      "another"
    ]
  }
}
*/
```

As for the name, "merge" seems natural but I would also consider using "copy" (as used in Go's [maps](https://pkg.go.dev/maps#Copy) package). Other possibility is to modify "concat" to support both arrays and maps.

eyupcanakman:
Adds a variadic `merge()` builtin that shallow-merges two or more maps into a new map. Later arguments override earlier ones (last-write-wins). Inputs are not modified.

merge({"a": 1, "b": 2}, {"b": 3, "c": 4})
// {"a": 1, "b": 3, "c": 4}

Keys keep their original type, so the result is `map[any]any`, the same type `fromPairs()` returns. The copy loop uses `MapRange` rather than `MapKeys` plus `MapIndex`, because `MapIndex` looks a key up by equality and so drops any key that is not equal to itself, such as NaN.

The new `cannot merge` runtime error is registered in the fuzz skip list, the same way #912 and #940 registered theirs.

`toJSON()` cannot serialize the result. `fromPairs()` and `groupBy()` output has the same problem today, since `encoding/json` rejects a map whose declared key type is `interface{}`.

Fixes #895

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -807,6 +807,46 @@ var Builtins = []*Function{
 		},
 	},
 	{
+		Name: "merge",
+		Safe: func(args ...any) (any, uint, error) {
+			if len(args) < 2 {
+				return nil, 0, fmt.Errorf("invalid number of arguments (expected at least 2, got %d)", len(args))
+			}
+
+			out := reflect.MakeMap(mapType)
+
+			for _, arg := range args {
+				v := reflect.ValueOf(arg)
+
+				if v.Kind() != reflect.Map {
+					return nil, 0, fmt.Errorf("cannot merge %s", v.Kind())
+				}
+
+				iter := v.MapRange()
+				for iter.Next() {
+					out.SetMapIndex(iter.Key(), iter.Value())
+				}
+			}
+
+			return out.Interface(), uint(out.Len()), nil
+		},
+		Validate: func(args []reflect.Type) (reflect.Type, error) {
+			if len(args) < 2 {
+				return anyType, fmt.Errorf("invalid number of arguments (expected at least 2, got %d)", len(args))
+			}
+
+			for _, arg := range args {
+				switch kind(arg) {
+				case reflect.Interface, reflect.Map:
+				default:
+					return anyType, fmt.Errorf("cannot merge %s", arg)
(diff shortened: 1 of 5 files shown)


builtin/builtin.go
```diff
@@ -1034,6 +1034,45 @@ var Builtins = []*Function{
...
+			if len(args) < 2 {
+				return nil, 0, fmt.Errorf("invalid number of arguments (expected at least 2, got %d)", len(args))
+			}
+
+			out := make(map[string]any)
+
+			for _, arg := range args {
+				v := reflect.ValueOf(arg)
+
+				if v.Kind() != reflect.Map {
```
sanmaxdev: `fmt.Sprint` collapses distinct map keys that have the same text form. For example, merging `map[int]string{1: "int"}` with `map[string]string{"1": "string"}` returns one `"1"` entry and silently loses the integer-keyed value. Existing map utilities preserve arbitrary key types. Please keep the original keys, for example in `map[any]any`, and add a mixed-key regression test.

sanmaxdev (changes requested):
One key-preservation issue needs fixing before this is ready.

builtin/builtin.go
eyupcanakman: Switched the output map to `map[any]any` so the original keys survive, and added a mixed-key regression test with `map[int]string{1: "int"}` and `map[string]string{"1": "string"}`. The copy uses `MapRange`, since looking each key back up with `MapIndex` silently drops a NaN key.

PR #949, opened 2026-03-27
feat: support DisableBuiltin("$env") to hide the $env variable

### LINKED ISSUE #710: [Request/Question] Allow control over if/how the $env var is exposed
I'm considering using `expr` at DAYJOB and would like to not expose the `$env` var using that name. We're planning on using at as a DSL and the domain has vars which conflict with `$env` and/or look similar enough that I want to stray from any confusion. 

I skimmed thru the code and it doesn't look like I can hide it from the expr programs via configuration today. Please call out if I just missed that. 

Questions I have for you:
- Does this var need to be exposed to end users? I can understand the convenience of it, but do you see it as a must have or a nice to have?
- Would you be amenable to allowing it to be exposed under a different name? If not, would you be amenable to allowing it to not be exposed (something like the DisableBuiltin() directive)? 

I'm happy to do the work to support my request but I wanted to ask before I went and put up a  PR that doesn't align with your direction for `expr`.

boinger:
## Summary

`$env` is a hardcoded special variable with custom handling in the checker and compiler. It's not in the `Builtins` map, so `DisableBuiltin("$env")` previously had no effect.

This PR checks `config.Disabled["$env"]` before each `$env` special-case handler. When disabled, `$env` falls through to normal identifier resolution, which errors in strict mode as `"unknown name $env"`. This is consistent with how `DisableBuiltin` works for regular builtins.

### Changes

- **`checker/checker.go`**: Guard `$env` handling in `identifierNode`, `memberNode`, `callNode`, and `callBuiltin` (the `get($env, ...)` path) with `!v.config.Disabled["$env"]`
- **`compiler/compiler.go`**: Guard `OpLoadEnv` emission in `IdentifierNode` with `!c.config.Disabled["$env"]`
- **`test/issues/710/issue_test.go`**: 7 subtests covering all disabled paths (standalone, member access, call, `get()`), non-strict mode behavior, and regression tests for default behavior

### Use case

Users embedding expr as a DSL may have domain variables that conflict with `$env`, or may want to restrict access to the full environment object for sandboxing.

Fixes #710

### DIFF
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -252,7 +252,7 @@ func (v *Checker) identifierNode(node *ast.IdentifierNode) Nature {
 			return v.varScopes[i].nature
 		}
 	}
-	if node.Value == "$env" {
+	if node.Value == "$env" && !v.config.Disabled["$env"] {
 		return Nature{}
 	}
 
@@ -514,7 +514,7 @@ func (v *Checker) chainNode(node *ast.ChainNode) Nature {
 
 func (v *Checker) memberNode(node *ast.MemberNode) Nature {
 	// $env variable
-	if an, ok := node.Node.(*ast.IdentifierNode); ok && an.Value == "$env" {
+	if an, ok := node.Node.(*ast.IdentifierNode); ok && an.Value == "$env" && !v.config.Disabled["$env"] {
 		if name, ok := node.Property.(*ast.StringNode); ok {
 			strict := v.config.Strict
 			if node.Optional {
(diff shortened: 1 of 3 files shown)


sanmaxdev (commented):
Checked the disabled `$env` path across the checker and compiler. In non-strict mode it now compiles as a normal unresolved identifier instead of emitting `OpLoadEnv`, while the default path remains unchanged.

Verified:
- `go test ./...`
- `go vet ./...`
- `go test -race -count=1 ./checker ./compiler ./test/issues/710`

antonmedv:
Will merge it soon.

PR #955, opened 2026-04-12
fix: support ConstExpr with functions defined via expr.Function

### LINKED ISSUE #809: expr.ConstExpr panics if function was defined with expr.Function instead of env
**Short description**: `expr.ConstExpr` requires that the given function was defined in the environment and does not take into account those functions defined with `expr.Function`.

**Why it matters**: This prevents usage of structs as environment in many cases, since it's impractical to define all the functions in the environment struct so we resort to add functionality with `expr.Function`.

**Workaround**: It is possible to directly define the signature in the configuration if we directly set the value for the function in the `expr/conf.(*Config).ConstFns`.

**Limitations**: a ConstExpr function can have a single signature because for each function name, we can only assign one value in `ConstFns`. Maybe worth creating a Feature Request issue?

<details><summary>Example (<a href="https://go.dev/play/p/szukoDzjRnx">Go Playground</a>)</summary><p>

```go
package main

import (
	"fmt"

	"github.com/expr-lang/expr"
)

type MyEnv struct {
	Value string
}

func identity(x any) any {
	return x
}

func main() {
	code := `[identity(1), identity(Value)]`

	options := []expr.Option{
		expr.Env(MyEnv{}),
		expr.Function(
			"identity",
				return identity(params[0]), nil
			},
			identity,
		),
		expr.ConstExpr("identity"), // Mark identity func as constant expression.
	}

	if err != nil {
		panic(err)
	}

	env := MyEnv{
		Value: "thing",
	}
	if err != nil {
		panic(err)
	}

	fmt.Println(output)

}
```

</p></details>

<details><summary>Workaround for single-signature (<a href="https://go.dev/play/p/CsJGTg1j6a7">Go Playground</a>)</summary><p>

```go
package main

import (
	"fmt"
	"reflect"

	"github.com/expr-lang/expr"
	"github.com/expr-lang/expr/conf"
)

type MyEnv struct {
	Value string
}

func identity(x any) any {
	return x
}

func main() {
	code := `[identity(1), identity(Value)]`

	options := []expr.Option{
		expr.Env(MyEnv{}),
		expr.Function(
			"identity",
				return identity(params[0]), nil
			},
			identity,
		),
		func(c *conf.Config) {
			c.ConstFns["identity"] = reflect.ValueOf(identity) // Mark identity func as constant expression.
		},
	}

	if err != nil {
		panic(err)
	}

	env := MyEnv{
		Value: "thing",
	}
	if err != nil {
		panic(err)
	}

	fmt.Println(output)

}
```

</p></details>

Yanhu007:
Fixes #809

## Problem

`expr.ConstExpr("fn")` panics when `fn` was defined using `expr.Function()` instead of being a method on the environment struct:

```go
expr.Function("identity", func(params ...any) (any, error) {
    return params[0], nil
}, new(func(any) any))

expr.Compile(`identity(1)`, expr.ConstExpr("identity")) // panic!
```

This is because `ConstExpr` only looks in the environment object via `runtime.Fetch`, not in the `Functions` table populated by `expr.Function()`.

## Fix

Check `c.Functions` first and extract the appropriate function implementation (`Func`, `Fast`, or `Safe`) before falling back to the environment object lookup.

### DIFF
--- a/conf/config.go
+++ b/conf/config.go
@@ -74,6 +74,22 @@ func (c *Config) WithEnv(env any) {
 }
 
 func (c *Config) ConstExpr(name string) {
+	// Check if the function is defined via expr.Function first.
+	if fn, ok := c.Functions[name]; ok {
+		if fn.Func != nil {
+			c.ConstFns[name] = reflect.ValueOf(fn.Func)
+			return
+		}
+		if fn.Fast != nil {
+			c.ConstFns[name] = reflect.ValueOf(fn.Fast)
+			return
+		}
+		if fn.Safe != nil {
+			c.ConstFns[name] = reflect.ValueOf(fn.Safe)
+			return
+		}
+		panic(fmt.Errorf("const expression %q must have a function implementation", name))
+	}
 	if c.EnvObject == nil {
 		panic("no environment is specified for ConstExpr()")
 	}


antonmedv:
Could you please also add a test to cover this API usage?

sanmaxdev (commented):
Verified the fix works. Registered a function via `expr.Function()` with `expr.ConstExpr()` and no environment, and constant folding produces the expected result (`double(21)` folds to `42`, `double(21) + 1` folds to `43`). All existing tests pass.

One gap: no test coverage for the new `c.Functions` lookup path. The existing `TestConstExpr_error_*` tests all go through the environment branch. A test that registers a function via `expr.Function()` and compiles with `expr.ConstExpr()` (without `expr.Env()`) would guard this path against regressions.

PR #958, opened 2026-05-11
feat: support uniqBy function

XieChengYi:
## Summary
- add uniqBy(array, predicate) as a predicate builtin
- compile uniqBy through the VM loop/accumulator path, preserving first item per unique key
- add safe hash fast path for common keys with runtime.Equal fallback for complex keys
- document uniqBy and cover it with builtin and public API tests

## Tests
- go test ./...

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -99,6 +99,11 @@ var Builtins = []*Function{
 		Types:     types(new(func([]any, func(any) any) map[any][]any)),
 	},
 	{
+		Name:      "uniqBy",
+		Predicate: true,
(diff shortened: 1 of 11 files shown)


XieChengYi:
Hi maintainers, thanks for the great project.

This PR adds `uniqBy(array, predicate)` as a natural extension of the existing `uniq` builtin, following the same predicate builtin pipeline as `groupBy` and `sortBy`.

I would appreciate any feedback on the API name, docs wording, or implementation approach. Happy to adjust it to better fit the project.
@antonmedv

PR #964, opened 2026-05-27
reject non-comparable keys in fromPairs

netliomax25-code:
fromPairs builds a map[any]any, so a pair whose key is a slice or
map hits a raw "hash of unhashable type" runtime panic. Check the
key is comparable before SetMapIndex and return an error instead,
matching the groupBy handling added in #940.

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -791,6 +791,9 @@ var Builtins = []*Function{
 				}
 				key := pair.Index(0)
 				value := pair.Index(1)
+				if k := key.Interface(); k != nil && !reflect.TypeOf(k).Comparable() {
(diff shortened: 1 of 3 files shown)


netliomax25-code:
gentle ping

sanmaxdev (commented):
Verified that `fromPairs` now returns a descriptive error for slice keys instead of panicking.

Checks passed:
- `go test ./...`
- `go test -race .`
- `GOARCH=386 go build`

netliomax25-code:
Good to see it holds up under the race and debug runs too. The fuzz skip pattern covers the new error string, so generated expressions that hit it are treated as an expected error rather than a failure.

PR #965, opened 2026-05-28
fix(builtin): reject negative count in take

netliomax25-code:
take([1,2,3], -1) surfaces a raw reflect.Value.Slice "slice index
out of bounds" runtime error. The negative count skips the > v.Len()
clamp and reaches v.Slice(0, -1) unmodified. Reject n < 0 the same
way repeat does and return a descriptive error.

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -657,6 +657,9 @@ var Builtins = []*Function{
 			if !n.CanInt() {
 				return nil, fmt.Errorf("cannot take %s elements", n.Kind())
 			}
+			if n.Int() < 0 {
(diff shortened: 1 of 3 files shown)


netliomax25-code:
gentle ping

netliomax25-code:
any update?

sanmaxdev (commented):
Checked the negative-count guard and its regression coverage. The focused error case, full suite, race check, and debug build all pass.

Passed:
- `go test ./...`
- `go test -race .`

netliomax25-code:
Thanks for running the suite and race check against this. I've left the branch as-is so your verification still matches the head commit, but I can rebase onto master whenever a maintainer wants it current.

PR #967, opened 2026-05-28
guard map key type before MapIndex in get

netliomax25-code:
The get builtin documents that it returns nil instead of panicking, but the map branch calls v.MapIndex(reflect.ValueOf(i)) directly. When the key is not assignable to the map's key type, or is an unhashable value reaching a map[any]any, reflect panics and the vm surfaces a raw reflect message like "reflect.Value.MapIndex: value of type ... is not assignable". Reachable from a dynamically typed map, e.g. get(m, [1,2]) where m has static type any. Check the key against the map key type and comparability first and fall back to nil, matching how missing keys already behave.

### DIFF
--- a/builtin/builtin_test.go
+++ b/builtin/builtin_test.go
@@ -300,6 +300,29 @@ func TestBuiltin_errors(t *testing.T) {
 	}
 }
 
+func TestGet_mapKeyTypeMismatch(t *testing.T) {
+	env := map[string]any{
+		"strMap": map[string]any{"a": 1},
+		"intMap": map[int]any{1: "x"},
+		"anyMap": map[any]any{"a": 1},
+	}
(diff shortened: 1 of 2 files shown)


netliomax25-code:
gentle ping

sanmaxdev (commented):
The guard covers both key-type mismatches and non-comparable interface keys while preserving valid map lookups.

Checked:
- `go test ./builtin`
- `go test ./...`
- `go test -race .`

netliomax25-code:
Matches what I see locally, `go test ./builtin` and `go test ./...` both pass on the current head. The guard only short-circuits when the key type can't legally index the map, so valid lookups are unaffected.

netliomax25-code:
friendly bump

PR #968, opened 2026-05-29
handle arbitrary slice types in sort builtin

netliomax25-code:
sort only special-cases []any/[]int/[]float64/[]string, so a dynamically typed slice like []int64 or []uint falls through the switch and returns an empty slice with no error, silently dropping every element. Handle any slice or array through reflection like reverse and uniq do.

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -998,12 +998,16 @@ var Builtins = []*Function{
 				}
 			default:
 				v := reflect.ValueOf(args[0])
-				if v.Kind() != reflect.Slice && v.Kind() != reflect.Array {
-					return nil, 0, fmt.Errorf("cannot sort %s", v.Kind())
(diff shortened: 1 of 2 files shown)


antonmedv (approved):

antonmedv:
looks like a lot of generated expressions now are falling

netliomax25-code:
Yeah, the error path was too broad. sort used to silently return empty for anything that wasn't one of the special-cased slice types, so the generated corpus has a bunch of expressions sorting maps/nil/struct slices that relied on that. Dropped the error and narrowed the reflection fallback to only pick up slices/arrays whose element kind is actually orderable (ints, uints, floats, string, bool). Everything else keeps the old no-op behavior. gen tests are green again.

netliomax25-code:
any update?

PR #969, opened 2026-05-29
return error on non-string element in join

netliomax25-code:
join([1,2]) hits arg.(string) on the []any branch and surfaces a raw "interface conversion: interface {} is int, not string" runtime error instead of the invalid argument for join error the other branches return. Check the assertion and return that error.

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -337,7 +337,11 @@ var Builtins = []*Function{
 			case []any:
 				var s []string
 				for _, arg := range args[0].([]any) {
-					s = append(s, arg.(string))
+					str, ok := arg.(string)
+					if !ok {
+						return nil, fmt.Errorf("invalid argument for join (type %s)", reflect.TypeOf(arg))
+					}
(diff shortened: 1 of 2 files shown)


antonmedv (approved):

builtin/builtin.go
```diff
@@ -996,6 +1000,15 @@ var Builtins = []*Function{
 				for i, v := range in {
 					array[i] = v
 				}
+			default:
+				v := reflect.ValueOf(args[0])
+				if v.Kind() != reflect.Slice && v.Kind() != reflect.Array {
+					return nil, 0, fmt.Errorf("cannot sort %s", v.Kind())
```
antonmedv: i guess this is from another pr

netliomax25-code: Right, that was from #968 leaking in because this branch was stacked on it. Rebased onto master and dropped the sort changes, so this PR is just the join fix now.

netliomax25-code:
gentle ping

PR #970, opened 2026-05-29
return error from date when only a location is given

netliomax25-code:
date(timezone("UTC")) strips the location then reads args[0] unchecked, panicking with a raw index-out-of-range; return the usual arg-count error instead.

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -517,6 +517,10 @@ var Builtins = []*Function{
 				args = args[1:]
 			}
 
+			if len(args) == 0 {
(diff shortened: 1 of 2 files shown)


netliomax25-code:
any update?

sanmaxdev (commented):
Checked the timezone-only `date()` path and its regression case. The guard returns the established argument-count error instead of indexing an empty slice.

Verified:
- `go test ./builtin`
- `go test ./...`

netliomax25-code:
Thanks for verifying. The branch is a bit behind master now but still merges cleanly, so it should be ready for a maintainer to take a look. Happy to rebase if that's preferred before merging.

PR #971, opened 2026-05-31
reject non-comparable keys in map index and in operator

netliomax25-code:
Indexing a map or testing membership with the `in` operator both reach `v.MapIndex(reflect.ValueOf(key))` in `runtime.Fetch`/`runtime.In`, which panics with a raw `hash of unhashable type` when the key is a slice or map; `groupBy(list,0)[reduce(list,B"")]` and `B"" in groupBy(list,0)` trigger it once the map/key is dynamically typed. Check `Comparable()` before the lookup, matching the groupBy key fix.


netliomax25-code:
any update?

sanmaxdev (approved):
The map-key guard covers both indexing and membership, and the regression cases exercise each path. Verified with `go test ./...` and `git diff --check origin/master...HEAD`.

PR #972, opened 2026-06-01
return error for non-comparable key in get builtin

netliomax25-code:
Noticed get indexes a map with reflect.Value.MapIndex(reflect.ValueOf(i)) but never checks the key is comparable. When the map key type is dynamic, e.g. groupBy returns map[any][]any, a slice key passes the checker and hits a raw "hash of unhashable type" panic at runtime, so get(groupBy([1,2,3], # % 2), [1]) blows up. Since get returns errors directly, return a clean "not comparable" error before the lookup, same as the groupBy fix.

### DIFF
--- a/builtin/builtin_test.go
+++ b/builtin/builtin_test.go
@@ -271,6 +271,7 @@ func TestBuiltin_errors(t *testing.T) {
 		{`date("error")`, `invalid date`},
 		{`get()`, `invalid number of arguments (expected 2, got 0)`},
 		{`get(1, 2)`, `type int does not support indexing`},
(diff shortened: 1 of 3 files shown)


netliomax25-code:
any update?

sanmaxdev (commented):
Checked the map-key guard and its regression coverage. The comparability check prevents the runtime panic while preserving the existing lookup path for valid keys.

Verified:
- `go test ./builtin ./test/fuzz`
- `go test ./...`

netliomax25-code:
Thanks for checking it over. Reran go test ./... on the branch today and it's still green; the branch is a bit behind master now but there are no conflicts, so it should rebase cleanly.

netliomax25-code:
gentle ping

PR #973, opened 2026-06-02
check struct field name is a string in Fetch and get

netliomax25-code:
runtime.Fetch (runtime.go:76) and the get builtin (lib.go:598) read a struct field name with an unchecked i.(string), so indexing a struct held in an any value with a non-string key panics with a raw interface conversion. Use a checked assertion so Fetch falls through to its usual "cannot fetch" error and get returns nil per its no-panic contract.

### DIFF
--- a/builtin/builtin_test.go
+++ b/builtin/builtin_test.go
@@ -211,6 +211,18 @@ func TestBuiltin(t *testing.T) {
 	}
 }
 
+func TestBuiltin_get_struct_non_string_key(t *testing.T) {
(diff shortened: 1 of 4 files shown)


netliomax25-code:
any update?

sanmaxdev (approved):
The struct-key guard is consistent across `get` and runtime field access, and both paths have regression coverage.

Passed:
- `go test ./...`
- `git diff origin/master...HEAD --check`

PR #974, opened 2026-06-02
fix(ast): avoid panic in ConstantNode.String on non-JSON floats

netliomax25-code:
ConstantNode.String panics when json.Marshal rejects the value, which happens for NaN and ±Inf. Constant folding produces these for array literals like [0/0], and predicateCombination calls String on the collection argument while optimizing, so any([0/0], # > 0) || any([0/0], # > 0) makes expr.Compile panic instead of returning a program. Compile does not recover, so the panic reaches the host. Fall back to fmt for values json cannot encode.

### DIFF
--- a/ast/print.go
+++ b/ast/print.go
@@ -43,7 +43,11 @@ func (n *ConstantNode) String() string {
 	}
 	b, err := json.Marshal(n.Value)
 	if err != nil {
-		panic(err)
+		// json.Marshal rejects values such as NaN and ±Inf, which can
+		// reach here after constant folding (e.g. an array literal like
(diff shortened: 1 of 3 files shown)


netliomax25-code:
any update?

sanmaxdev (commented):
I checked the non-JSON float path through `ConstantNode.String()` and the optimizer regression cases. The fallback keeps the existing JSON output for marshalable constants while preventing the NaN/Inf panic path.

Verified locally:
- `go test ./ast ./optimizer`
- `go test ./...`
- `git diff --check`

PR #983, opened 2026-07-26
Match float literals to float32 comparisons

### LINKED ISSUE #609: Floating number comparison returns incorrect result
When you try to compare floating number with floating literal like `Number == 12.34`, it returns false even when `Number` is 12.34

Below is code to reproduce
https://go.dev/play/p/FFgPkXjIURm
```
package main

import (
	"fmt"

	"github.com/expr-lang/expr"
)

type Env struct {
	Number float32
}

func main() {
	code := `Number == 12.34`

	program, err := expr.Compile(code, expr.Env(Env{}))
	if err != nil {
		panic(err)
	}

	env := Env{
		Number: 12.34,
	}
	if err != nil {
		panic(err)
	}

	fmt.Println(output)
}
```

deepakganesh78:
Fixes #609

## Root cause

Expr compiles floating-point literals as `float64`. Comparing one directly with a `float32` operand therefore compares the rounded `float32` value with the original `float64` literal, unlike Go's conversion of the untyped literal to `float32`.

## Fix

After type checking, patch float literals used directly in comparisons with `float32` operands into typed `float32` constants. This handles either operand order and signed literals without changing mixed `float32`/`float64` arithmetic.

## Tests

- `go test .\checker . -run 'Test(Float32LiteralComparisons|Check)' -count=1`
- `go test ./...`
- `go vet ./...`

The regression test covers optimized and unoptimized compilation, both operand orders, signed literals, relational and inequality comparisons, and verifies `float64` behavior remains unchanged.

`go test -race .` was not available on this Windows host because the installed Go toolchain has CGO disabled.

### DIFF
--- a/ast/node.go
+++ b/ast/node.go
@@ -112,8 +112,8 @@ type BytesNode struct {
 
 // ConstantNode represents a constant.
 // Constants are predefined values like nil, true, false, array, map, etc.
-// The parser.Parse will never generate ConstantNode, it is only generated
-// by the optimizer.
+// The parser.Parse will never generate ConstantNode; it is generated after
+// parsing by optimizers and patchers.
 type ConstantNode struct {
 	base
 	Value any // Value of the constant.
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -78,7 +78,13 @@ func (v *Checker) PatchAndCheck(tree *parser.Tree, config *conf.Config) (reflect
 		// Run patchers that require multiple passes next (currently only Operator patching)
 		v.runVisitors(tree, true)
 	}
-	return v.Check(tree, config)
+
+	typ, err := v.Check(tree, config)
+	if err != nil {
+		return typ, err
+	}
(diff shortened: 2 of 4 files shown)


sanmaxdev (commented):
Verified this fixes #609. Reproduced `F32 == 12.34` returning false on master, and it returns true on this branch. The patcher approach is what was requested in the issue thread, and it correctly handles both operand orders plus signed literals without touching mixed float32/float64 arithmetic.

Verification run on the PR head:
- `go test ./...` (full suite, no failures)
- `go vet ./...` (clean)
- `go test -race -count=1 ./checker .` (green)

One test gap: `isComparisonOperator` returns true for `<`, `>`, and `>=`, but the table in `TestFloat32LiteralComparisons` only exercises `==`, `!=`, and `<=`. Adding rows for the remaining three would cover every branch of the helper and guard against a future regression that narrows the accepted operator set.

PR #986, opened 2026-08-18
feat: add MockNow option for deterministic now() in tests

### LINKED ISSUE #754: Allow mockable/predictable now()
Its hard to test out programs that rely on time for example:
```
now().Sub(StartTime) >= 40*duration("24h")
```
Would it be possible to allow a testing option for `now()` to give a predictable/settable time?

ChrisJr404:
Fixes #754.

`now()` makes expressions hard to test because it returns a different value on every run. As suggested in the issue, this adds a small patcher (following the `WithTimezone` pattern) plus a public option so `now()` can be pinned to a fixed time.

### Usage
```go
program, err := expr.Compile(`now().Sub(StartTime) >= 40*duration("24h")`,
    expr.Env(env),
    expr.MockNow(time.Date(2024, 5, 7, 23, 0, 0, 0, time.UTC)),
)
```
The compiled program now always sees that fixed time for `now()`, so tests are deterministic. No boilerplate patcher and no need to `DisableBuiltin("now")` + re-register a function.

### What's included
- `patcher.WithNow` — replaces `now()` builtin calls with a constant time node while patching the AST.
- `expr.MockNow(t time.Time)` — the public option wiring the patcher, mirroring `expr.Timezone`.
- It composes with `Timezone`: if a location was injected into `now()`, the fixed time is returned in that location.
- Docs section under `docs/configuration.md`.

### Notes
- Purely additive; no existing builtins or tests changed, default behavior unchanged.
- Tests cover the basic case, a deterministic derived value (`now().Year()`), and the `Timezone` + `MockNow` combination.
- `go build`, `go vet`, and `go test ./ ./patcher/ ./builtin/ ./checker/` pass. (`test/examples` fails to build on a clean `master` too — unrelated to this change.)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -132,3 +132,17 @@ The timezone is used for the following functions:
 date("2024-11-23 12:00:00") // parses the date in the specified timezone
 now() // returns the current time in the specified timezone
 ```
+
+## Mock now()
+
+Expressions that depend on `now()` are hard to test because the result changes
+on every run. The [`MockNow`](https://pkg.go.dev/github.com/expr-lang/expr#MockNow)
+option replaces `now()` with a fixed time, so the compiled program always
+returns the same value.
+
+```go
+program, err := expr.Compile(code, expr.MockNow(time.Date(2024, 11, 23, 12, 0, 0, 0, time.UTC)))
+```
+
+`MockNow` can be combined with [`Timezone`](#timezone) to return the fixed time
+in the specified timezone.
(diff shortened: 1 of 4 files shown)


sanmaxdev (commented):
The composition is order-dependent. `MockNow` replaces `now()` with a `ConstantNode`, so a later `Timezone` visitor has no `BuiltinNode` to modify. `expr.Compile(code, expr.MockNow(now), expr.Timezone("Asia/Kamchatka"))` therefore keeps the original location even though the docs say the options can be combined.

Please make both option orders work, or document the required order and add a regression test for it.

